PR #3, opened 2018-02-18
Passing branch name to spinmint scripts

jespino:
This is a proposal, and have some extra work in the spinmint configuration (another PR already done) and the jenkins configuration (Not done yet).


jespino:
I think it can be merged anyway, only add the branch information to the spinmint without changing any behavior.

PR #5, opened 2018-05-11
Add ability to deploy a test server and test the upgrade

cpanato:
Next stage is to refactor to use terraform. this is a quick and small win for now

### DIFF
--- a/server/spinmint.go
+++ b/server/spinmint.go
@@ -118,7 +118,7 @@ func waitForBuildAndSetupLoadtest(pr *model.PullRequest) {
 	commentOnIssue(pr.RepoOwner, pr.RepoName, pr.Number, results.String())
 }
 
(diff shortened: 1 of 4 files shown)


server/spinmint.go
```diff
@@ -173,6 +173,61 @@ func waitForBuildAndSetupSpinmint(pr *model.PullRequest) {
 	commentOnIssue(pr.RepoOwner, pr.RepoName, pr.Number, message)
 }
 
+func waitForBuildAndSetupSpinmintUpgrade(pr *model.PullRequest) {
```
crspeller: Can we reuse this code instead of copy paste and just parameterize this function?

cpanato: yep, will change

cpanato: done

crspeller:
@cpanato Merge when ready.

PR #22, opened 2019-02-22
[MM-11148] Convert mattermod to use structured logging #10315

fcorrea:
#### Summary
This PR drops the use of l4g (unmaintained) and replace it by mlog, which is Mattermost wrapper around Uber's zap logging lib.

The original ticket suggested to include a CLI to pull or monitor logs or factor out mlog out of Mattermost code base.
I'm more inclined for factoring out mlog and could even lead the effort if necessary. IMO having a few CLI for mlog would be best.

#### Ticket Link
[MM-11148](https://mattermost.atlassian.net/browse/MM-11148)

### DIFF
--- a/config/config-mattermod.json
+++ b/config/config-mattermod.json
@@ -74,5 +74,18 @@
     "AWSDnsSuffix": "",
 
     "MattermostWebhookURL": "",
-    "MattermostWebhookFooter": ""
+    "MattermostWebhookFooter": "",
+
+    "LogSettings": {
+        "EnableConsole": true,
+        "ConsoleLevel": "DEBUG",
+        "ConsoleJson": true,
+        "EnableFile": true,
+        "FileLevel": "INFO",
+        "FileJson": true,
+        "FileFormat": "",
+        "FileLocation": "",
+        "EnableWebhookDebugging": true,
+        "EnableDiagnostics": true
+    },
 }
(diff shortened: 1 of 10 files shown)


fcorrea:
Hey @crspeller ! Thanks for taking a look and clarifying.

I understand it better now and I'll make the changes to reflect that today. 
 
Thanks!

> Thanks for the effort!
> The idea of converting to using structured logging is to allow for easy parsing of useful information for a log aggregator. Using fmt.Sprintf() to create messages is not really any better than the current system. What we would want here is to use the mlog API as intended to split out useful information into fields.
> For instance:
> 
> ```go
> mlog.Info(fmt.Sprintf("Commenting on issue %v Comment: %v", strconv.Itoa(number), comment))
> ```
> 
> would be better as:
> 
> ```go
> mlog.Info("Commenting on issue", mlog.Int("issue", number), mlog.String("comment", comment))
> ```

fcorrea:
Hi @crspeller . 

This one is ready for one more review pass.
I deliberately created/added mlog identifiers as I went changing it but didn't follow any naming convensions. Please, let me know if they work or if you want them changed.

Thanks!

fcorrea:
> Looks better!
> BTW there is mlog.Err for errors instead of making them strings. Not going to block on that change though.

Thanks @crspeller for pointing that out. I've just addressed that and now we're using mlog.Err as suggested.

cpanato:
i will merge tomorrow and deploy 😄

PR #24, opened 2019-03-01
add lint

cpanato:
Adding inital version of the makefile and add at least lint to check prs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,8 +4,8 @@ DEP ?= $(shell command -v dep 2> /dev/null)
 PACKAGES=$(shell go list ./...)
 
 ## Checks the code style, tests, builds and bundles the plugin.
+.PHONY: all
 all: check-style
-
 
 ## Runs govet and gofmt against all packages.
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -0,0 +1,45 @@
+GO ?= $(shell command -v go 2> /dev/null)
+DEP ?= $(shell command -v dep 2> /dev/null)
+
+PACKAGES=$(shell go list ./...)
+
+## Checks the code style, tests, builds and bundles the plugin.
+all: check-style
```
hmhealey: This needs a `.PHONY`

Makefile
```diff
@@ -0,0 +1,45 @@
...
+
```
hmhealey: There's an extra blank line here. Maybe we need a lint for Makefiles 😛

Makefile
cpanato: done

Makefile
cpanato: done

PR #30, opened 2019-07-10
Use mattermost-cloud request objects

gabrieljackson:
This change moves mattermod over to using copies of the structs
that are used by the provisioning server in the mattermost-cloud
repo. This helps make mattermod less brittle when interacting with
provisioning servers.

This change also cleans up some logging and messages sent to PRs.

### DIFF
--- a/server/pull_request.go
+++ b/server/pull_request.go
@@ -33,11 +33,11 @@ func handlePullRequestEvent(event *PullRequestEvent) {
 
 			commentOnIssue(pr.RepoOwner, pr.RepoName, pr.Number, Config.DestroyedSpinmintMessage)
 			go destroySpinmint(pr, spinmint.InstanceId)
-			go destroySpinmintExperimental(pr, spinmint.InstanceId)
+			go destroySpinWick(pr, spinmint.InstanceId)
 		}
 	} else if event.Action == "synchronize" {
 		checkCLA(pr)
-		upgradeTestServer(pr)
+		updateSpinWick(pr)
 	}
 
 	checkPullRequestForChanges(pr)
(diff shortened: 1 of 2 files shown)


server/pull_request.go
```diff
@@ -176,9 +176,9 @@ func handlePRLabeled(pr *model.PullRequest, addedLabel string) {
 		commentOnIssue(pr.RepoOwner, pr.RepoName, pr.Number, Config.StartLoadtestMessage)
 		go waitForBuildAndSetupLoadtest(pr)
 	} else if addedLabel == Config.SetupSpinmintExperimentalTag && !messageByUserContains(comments, Config.Username, "Mattermost test server created!") {
```
stylianosrigas: @gabrieljackson Will we change the label not to be called experiementaltag?

server/test_servers_experimental.go
```diff
@@ -160,16 +199,15 @@ func setupSpinmintExperimental(pr *model.PullRequest) (string, error) {
...
 
-	return createInstallationRequest.ID, nil
+	return installationID, nil
 }
 
-func upgradeTestServer(pr *model.PullRequest) {
-
+func updateSpinWick(pr *model.PullRequest) {
 	for _, label := range pr.Labels {
 		if label == Config.SetupSpinmintExperimentalTag {
```
stylianosrigas: @gabrieljackson Same here

gabrieljackson:
Thanks @stylianosrigas. We will want to change the tags as you mentioned, but we are not quite ready for that. That should happen in a follow-up PR.

stylianosrigas:
@gabrieljackson All good then, feel free to resolve my comments ;)

PR #31, opened 2019-07-11
Add HA SpinWick label option

gabrieljackson:
This allows for mattermod to request different installation sizes
from the provisioner.


gabrieljackson:
Note that this changes the config which will require manual updating on mattermod.

PR #33, opened 2019-07-15
Show SpinWick test user data in a table

gabrieljackson:
Because that is fancier.


gabrieljackson:
I think I got the formatting right, but I am unable to directly test. It should look like:

| Account Type | Username | Password |
|---|---|---|
| Admin | sysadmin | Sys@dmin123 |
| User | user-1 | User-1@123 |

cpanato:
merging

PR #39, opened 2019-08-01
[MM-17073] Mattermod Refactor #2

gabrieljackson:
The intention of this PR is to remove mattermod tech debt in order to 
provide a foundation for a better SpinWick experience. The primary
goal of this work was to begin addressing some of the race conditions
that exist when working with SpinWicks.

The following was included in these commits: 

- Enhanced Server struct and changed all related functions to
   methods of Server. The Server object is no longer a shared
   variable.
 - Moved the server config to belong to the server struct and not
   be a shared variable.
 - Provided a way to cleanly create, start, and stop a server.
 - Resolved issues with not returning errors when we should be.
- Refactors the spinwick code to no longer require the mattermod
   database.
 - Cleans up logic for creating, updating, and deleting SpinWicks
   to primarily occur as a direct result of github events.
 - Adds more tests where possible.

https://mattermost.atlassian.net/browse/MM-17073

### DIFF
--- a/cloud/installation.go
+++ b/cloud/installation.go
@@ -1,0 +1,90 @@
+package cloud
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+
+	"github.com/pkg/errors"
+)
+
+// GetInstallation returns an installation for a given ID.
+func GetInstallation(serverURL, installationID string) (Installation, error) {
+	var installation Installation
+	url := fmt.Sprintf("%s/api/installation/%s", serverURL, installationID)
+	resp, err := makeRequest("GET", url, nil)
+	if err != nil {
+		return installation, errors.Wrap(err, "unable to get installation")
+	}
+	defer resp.Body.Close()
+
+	err = json.NewDecoder(resp.Body).Decode(&installation)
+	if err != nil && err != io.EOF {
+		return installation, errors.Wrap(err, "Error decoding installation")
+	}
+
+	return installation, nil
+}
+
+// GetInstallationList returns a list on non-deleted installations from the
+// provisioning server.
+func GetInstallationList(serverURL string) ([]Installation, error) {
+	var installations []Installation
+
+	url := fmt.Sprintf("%s/api/installations", serverURL)
+	resp, err := makeRequest("GET", url, nil)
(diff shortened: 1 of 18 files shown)


gabrieljackson:
@cpanato, I could use some guidance on how to test this before we merge it and push it into production. I can build and run it locally, but I am unsure how I properly get jenkins builds and/or which VPNs will need to be active to properly contact the provisioning server, CI/CD, and github at the same time.

cpanato:
@gabrieljackson you need to:
-  setup a jenkins instance or comment out the checks
- setup a repo that points to your local mattermod

cloud/installation.go
```diff
@@ -0,0 +1,90 @@
...
+
+	err = json.NewDecoder(resp.Body).Decode(&installations)
+	if err != nil && err != io.EOF {
+		return installations, errors.Wrap(err, "error decoding installations")
+	}
+
+	return installations, nil
+}
+
+// GetInstallationIDFromOwnerID returns the ID of an installtion that matches
```
cpanato: s/installtion/installation

server/pull_request.go
```diff
@@ -14,16 +14,52 @@ import (
...
-	if event.Action == "closed" {
-		go checkIfNeedCherryPick(pr)
-		if result := <-Srv.Store.Spinmint().Get(pr.Number, pr.RepoName); result.Err != nil {
+	switch event.Action {
+	case "labeled":
+		if event.Label == nil {
+			mlog.Error("Label event received, but label object was empty")
+			return
+		}
+		if s.isSpinWickLabel(*event.Label.Name) {
```
cpanato: dont we need to add the old test servers as well?

gabrieljackson: As we discussed in Mattermost, we will need to cleanup the mattermod DB after this is merged, but that can be done anytime without issue.

gabrieljackson:
I have completed some basic testing of this PR locally. There is a bunch of code I can't reasonably get to without a real build and new image to test, but the basic event logic is working.

At this point, I would request that we review and merge Monday so we can handle any issues that may arise.

cpanato:
I will deploy monday morning and will perform some prod tests

PR #40, opened 2019-08-05
uses new sizes for test servers

cpanato:


cpanato:
waiting we bump the operator and tag a new cloud as well

PR #41, opened 2019-08-07
remove label when test server fail and do some comments cleanup

cpanato:
- when a spinwick fail remove the label and the comments when requesting for the first time

### DIFF
--- a/server/spinwick.go
+++ b/server/spinwick.go
@@ -34,7 +34,7 @@ func (s *Server) handleCreateSpinWick(pr *model.PullRequest, size string) {
 			s.removeOldComments(comments, pr)
 		}
 		for _, label := range pr.Labels {
-			if label == s.Config.SetupSpinWick || label == s.Config.SetupSpinWickHA {
+			if s.isSpinWickLabel(label) {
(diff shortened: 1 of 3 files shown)


server/spinwick.go
```diff
@@ -27,7 +27,19 @@ func (s *Server) handleCreateSpinWick(pr *model.PullRequest, size string) {
...
 	if err != nil {
 		mlog.Error("Failed to create SpinWick", mlog.Err(err), mlog.String("repo_name", pr.RepoName), mlog.Int("pr", pr.Number), mlog.String("installation_id", installationID))
+		comments, err := s.getComments(pr.RepoOwner, pr.RepoName, pr.Number)
+		if err != nil {
+			mlog.Error("Error getting comments", mlog.Err(err))
+		} else {
+			s.removeOldComments(comments, pr)
+		}
+		for _, label := range pr.Labels {
+			if label == s.Config.SetupSpinWick || label == s.Config.SetupSpinWickHA {
```
gabrieljackson:
```suggestion
			if s.isSpinWickLabel(label) {
```

cpanato: thanks!

PR #43, opened 2019-08-12
add sha in the update server message and remove old comments

cpanato:
- add a sha in the comment when we update the test server
- remove old update test server messages to avoid confusion and make the PR a bit cleaner.

### DIFF
--- a/server/spinwick.go
+++ b/server/spinwick.go
@@ -188,7 +188,7 @@ func (s *Server) updateSpinWick(pr *model.PullRequest) (string, bool, error) {
 		return installationID, true, errors.Wrap(err, "error waiting for build link")
 	}
 
-	// Remove old messate to reduce the amount of similar messages and avoid confusion
+	// Remove old message to reduce the amount of similar messages and avoid confusion
 	serverNewCommitMessages := []string{
 		"New commit detected.",
 	}
--- a/server/spinwick.go
+++ b/server/spinwick.go
@@ -196,7 +196,7 @@ func (s *Server) updateSpinWick(pr *model.PullRequest) (string, bool, error) {
 	if errComments != nil {
 		mlog.Error("pr_error", mlog.Err(err))
 	} else {
-		s.removeServerUpdateOldComments(comments, serverNewCommitMessages, pr)
+		s.removeCommentsWithSpecificMessages(comments, serverNewCommitMessages, pr)
 	}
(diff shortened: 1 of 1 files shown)


server/spinwick.go
```diff
@@ -187,6 +188,16 @@ func (s *Server) updateSpinWick(pr *model.PullRequest) (string, bool, error) {
 		return installationID, true, errors.Wrap(err, "error waiting for build link")
 	}
 
+	// Remove old messate to reduce the amount of similar messages and avoid confusion
```
gabrieljackson:
```suggestion
	// Remove old message to reduce the amount of similar messages and avoid confusion
```

server/spinwick.go
```diff
@@ -226,8 +237,16 @@ func (s *Server) updateSpinWick(pr *model.PullRequest) (string, bool, error) {
 		return installationID, true, errors.Wrap(err, "encountered error waiting for installation to become stable")
 	}
 
+	// Remove old messate to reduce the amount of similar messages and avoid confusion
```
gabrieljackson:
```suggestion
	// Remove old message to reduce the amount of similar messages and avoid confusion
```

server/spinwick.go
```diff
@@ -226,8 +237,16 @@ func (s *Server) updateSpinWick(pr *model.PullRequest) (string, bool, error) {
...
+	if errComments == nil {
```
gabrieljackson: It looks like we should be running `s.getComments()` before this?

server/spinwick.go
```diff
@@ -606,3 +625,21 @@ func (s *Server) isSpinWickLabelInLabels(labels []string) bool {
 
 	return false
 }
+
+func (s *Server) removeServerUpdateOldComments(comments []*github.IssueComment, serverMessages []string, pr *model.PullRequest) {
```
gabrieljackson: This function name is a little confusing to me. Maybe something like `removeCommentsWithSpecificMessages`?

server/spinwick.go
cpanato: we get the comments in the beginning which will be the same and also we dont use the request against github, we save the token :)

server/spinwick.go
cpanato: sounds good

server/spinwick.go
gabrieljackson: Okay!

PR #44, opened 2019-08-13
fix upgrade installation when is in ha mode

cpanato:
when upgrading and the request was an HA server we were not setting the license causing the deletion of the license in the cluster

### DIFF
--- a/server/pull_request.go
+++ b/server/pull_request.go
@@ -54,7 +54,7 @@ func (s *Server) handlePullRequestEvent(event *PullRequestEvent) {
 		mlog.Info("PR has a new commit", mlog.String("repo", pr.RepoName), mlog.Int("pr", pr.Number))
 		s.checkCLA(pr)
 		if s.isSpinWickLabelInLabels(pr.Labels) {
-			mlog.Info("Pr have the spinWick label, starting the upgrade", mlog.String("repo", pr.RepoName), mlog.Int("pr", pr.Number))
+			mlog.Info("PR has a SpinWick label, starting upgrade", mlog.String("repo", pr.RepoName), mlog.Int("pr", pr.Number))
 			if s.isSpinWickHALabel(pr.Labels) {
 				s.handleUpdateSpinWick(pr, true)
 			} else {
--- a/server/pull_request.go
+++ b/server/pull_request.go
@@ -51,10 +51,16 @@ func (s *Server) handlePullRequestEvent(event *PullRequestEvent) {
 			s.handleDestroySpinWick(pr)
 		}
 	case "synchronize":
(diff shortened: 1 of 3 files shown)


server/pull_request.go
```diff
@@ -51,10 +51,16 @@ func (s *Server) handlePullRequestEvent(event *PullRequestEvent) {
 			s.handleDestroySpinWick(pr)
 		}
 	case "synchronize":
-		mlog.Info("PR has a new commit", mlog.String("repo", *event.Repo.Name), mlog.Int("pr", event.PRNumber))
+		mlog.Info("PR has a new commit", mlog.String("repo", pr.RepoName), mlog.Int("pr", pr.Number))
 		s.checkCLA(pr)
 		if s.isSpinWickLabelInLabels(pr.Labels) {
-			s.handleUpdateSpinWick(pr)
+			mlog.Info("Pr have the spinWick label, starting the upgrade", mlog.String("repo", pr.RepoName), mlog.Int("pr", pr.Number))
```
gabrieljackson:
```suggestion
			mlog.Info("PR has a SpinWick label, starting upgrade", mlog.String("repo", pr.RepoName), mlog.Int("pr", pr.Number))
```

server/spinwick.go
```diff
@@ -142,8 +142,8 @@ func (s *Server) createSpinWick(pr *model.PullRequest, size string, withLicense
 	return installationID, false, nil
 }
 
-func (s *Server) handleUpdateSpinWick(pr *model.PullRequest) {
-	installationID, sendMattermostLog, err := s.updateSpinWick(pr)
+func (s *Server) handleUpdateSpinWick(pr *model.PullRequest, withLicense bool) {
```
gabrieljackson: Is there a reason we don't want to just give every spinwick a license? I was wondering this earlier, but now I am thinking it may be best to keep things simple and allow for non-HA spinwicks to test other license features.

cpanato: @gabrieljackson for non ha, you might want to test without license and then later add a license to test other things. but i'm 0/5 to add a license to all

@jwilander any comments?

jwilander: I'm fine with giving them all a license

jwilander: 0/5 though, we can do that in a later PR

gabrieljackson: Same, only like 1/5. I was just thinking it would keep things simple in the mattermod code while we work on the other bugs.

gabrieljackson: Same, only like 1/5. I was just thinking it would keep things simple in the mattermod code while we work on the other bugs.

gabrieljackson: Same, only like 1/5. I was just thinking it would keep things simple in the mattermod code while we work on the other bugs.

cpanato: i will update

PR #47, opened 2019-08-26
MM-17925: Implement github ping event

metanerd:
https://github.com/mattermost/mattermost-server/issues/11921
https://mattermost.atlassian.net/browse/MM-17925
github_ping

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -187,7 +187,11 @@ func (s *Server) githubEvent(w http.ResponseWriter, r *http.Request) {
 
 	buf, _ := ioutil.ReadAll(r.Body)
 
-	pingEvent := PingEventFromJson(ioutil.NopCloser(bytes.NewBuffer(buf)))
+	var pingEvent *github.PingEvent
+	if r.Header.Get("X-GitHub-Event") == "ping" {
+		pingEvent = PingEventFromJson(ioutil.NopCloser(bytes.NewBuffer(buf)))
+	}
+
 	event := PullRequestEventFromJson(ioutil.NopCloser(bytes.NewBuffer(buf)))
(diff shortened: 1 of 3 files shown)


server/server.go
```diff
@@ -211,6 +212,11 @@ func (s *Server) githubEvent(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if pingEvent != nil {
```
cpanato: If you keep this check here the `s.handleIssueEvent(event)` will never run again.

you need to check if the header of the request contains the `X-GitHub-Event: Ping` as well

the ping event only happens once, when you create the webhook. after that this event will not be trigger

metanerd: Thank you for finding this one. 
Now the server throws a nullpointer exception in https://github.com/mattermost/mattermost-mattermod/blob/3e19389a41ff7a4c2d13e618a8d187a5083832cb/server/issue.go#L20 while trying to split the HTMLURL:
`parts := strings.Split(*event.Issue.HTMLURL, "/")`

I guess it is a local config problem. Anyway the `s.handleIssueEvent(event)` is reachable again now.

cpanato: just add this guard here: https://github.com/mattermost/mattermost-mattermod/blob/3e19389a41ff7a4c2d13e618a8d187a5083832cb/server/issue.go#L17
```
if event == nil || event.Issue == nil {
		return
	}
```
tested locally and it fix the panic
this is because when you setup the webhook you set the event pushes to be sent and this event there is no Issue data in that. but this is a good check to make it.
thanks so much for this PR :)

metanerd: Added, thank you.

cpanato:
@metanerd thanks so much for this addition! 🎉

PR #48, opened 2019-09-02
MM-17696: Change Mattermod to check Docker Registry instead of Jenkins Build

ankitrgadiya:
In this pull request, I replaced the Jenkins build check by polling on Docker registry instead. This way Spinwick can be spinned up as soon as the docker image with specific tag for the PR is pushed to Dockerhub. This, if coupled with moving the docker build before testing, will reduce the wait time for Spinwick.

Partially resolves [#11855](https://github.com/mattermost/mattermost-server/issues/11855)


server/builds.go
```diff
@@ -27,6 +32,25 @@ func (b *Builds) getInstallationVersion(pr *model.PullRequest) string {
 	return pr.Sha[0:7]
 }
 
+func (b *Builds) dockerRegistryClient(s *Server) (reg *registry.Registry, err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			err = errors.Errorf("panicked while connecting to docker registry: %s", r)
+		}
```
jwilander: This is interesting, does `registry` panic on failed connection often?

gabrieljackson: Yeah, I would also like to know the story here.

ankitrgadiya: While testing the first thing I checked was incorrect URL (without http part) and it panicked so I thought of catching that and return the error although I'm checking if the url is of correct format.

gabrieljackson: Hmm, interesting. I would prefer to remove the panic recovery here as it may hide other issues in the future. If we check the URL before using it in the registry, we should avoid this issue entirely anyway.

Thoughts, @jwilander?

jwilander: I'm ok with that 👍

ankitrgadiya: Yes, that makes sense. I'll make the change. Thanks

server/builds.go
```diff
@@ -45,6 +69,43 @@ func (b *Builds) buildJenkinsClient(s *Server, pr *model.PullRequest) (*Reposito
 	return repo, client, nil
 }
 
+func (b *Builds) waitForImage(ctx context.Context, s *Server, reg *registry.Registry, pr *model.PullRequest) (*model.PullRequest, error) {
+	for {
+		select {
+		case <-ctx.Done():
+			return pr, errors.New("timed out waiting for image to publish")
+		case <-time.After(30 * time.Second):
```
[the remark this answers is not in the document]
ankitrgadiya: I'll change it to 10. Thanks

PR #52, opened 2019-09-12
Add milestone to cherry-pick PR

hanzei:
I haven't tested this. Please double check.

### DIFF
--- a/server/cherry_pick.go
+++ b/server/cherry_pick.go
@@ -41,7 +41,7 @@ func (s *Server) handleCherryPick(eventIssueComment IssueComment) {
 		return
 	}
 
-	cmdOut, err := s.doCherryPick(args[1], pr)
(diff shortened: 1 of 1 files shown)


server/cherry_pick.go
```diff
@@ -41,7 +41,7 @@ func (s *Server) handleCherryPick(eventIssueComment IssueComment) {
 		return
 	}
 
-	cmdOut, err := s.doCherryPick(args[1], pr)
+	cmdOut, err := s.doCherryPick(args[1], nil, pr)
```
hanzei: Getting the code for this case do is actually quite tricky. And because it's a rare case, I opted to not do this for now.

hanzei:
@cpanato I would be great if you could review this PR this week, so we can use it for the cherry picks for 5.16.

PR #53, opened 2019-09-12
Move to go modules

hanzei:


server/issue.go
```diff
@@ -8,7 +8,7 @@ import (
 	"strings"
 	"time"
 
-	"github.com/google/go-github/github"
+	"github.com/google/go-github/v28/github"
```
cpanato: this is to lock the version?

hanzei: Kind of. This is just how imports work with go modules. The version _has_ to be part of the module path. See https://github.com/golang/go/wiki/Modules#semantic-import-versioning and https://github.com/google/go-github/blob/master/go.mod#L1

cpanato: 👍

hanzei:
Follow up question: Do we still want to use the `vendor` folder? If so, I would alter the makefile to always use it.

PR #56, opened 2019-10-07
Fix permission issues from PR/Forks triggering the circleci job with a org member

cpanato:
Context:
we have circleci for our CI pipelines for almost all public repos we have.
In circleci we use the feature called `context` which inject environment variables in the build, this env var contains some secrets to upload data to s3 or push docker images. This does not work when the build is running from a Fork/PR.
This PR fixes this issue we have that every time someone from the org needs to retrigger the job if that comes from a PR/Fork
this retriggers the job and the previous one will be canceled.

next: add a slash command to retrigger the build if that fail

### DIFF
--- a/server/circleci.go
+++ b/server/circleci.go
@@ -14,7 +14,7 @@ import (
 )
 
 func (s *Server) triggerCircleCiIfNeeded(pr *model.PullRequest) {
-	client := &circleci.Client{Token: s.Config.CircleCiToken}
+	client := &circleci.Client{Token: s.Config.CircleCIToken}
 
 	if strings.Contains(pr.FullName, "mattermost/") {
 		// It is from upstream mattermost repo dont need to trigger the circleci because org members
--- a/server/circleci.go
+++ b/server/circleci.go
@@ -1,0 +1,46 @@
+// Copyright (c) 2017-present Mattermost, Inc. All Rights Reserved.
+// See License.txt for license information.
+
+package server
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/mattermost/mattermost-mattermod/model"
+	"github.com/mattermost/mattermost-server/mlog"
+
+	"github.com/cpanato/go-circleci"
+)
+
+func (s *Server) triggerCircleCiIfNeeded(pr *model.PullRequest) {
+	client := &circleci.Client{Token: s.Config.CircleCiToken}
+
+	if strings.Contains(pr.FullName, "mattermost/") {
+		// It is from upstream mattermost repo dont need to trigger the circleci because org members
+		// have permissions
+		return
+	}
+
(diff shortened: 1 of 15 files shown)


server/circleci.go
```diff
@@ -0,0 +1,46 @@
...
+
+	// Checking if the repo have circleci setup
+	builds, err := client.ListRecentBuildsForProject(pr.RepoOwner, pr.RepoName, "master", "", 5, 0)
+	if err != nil {
+		mlog.Error("listing the circleci project", mlog.String("repo", pr.RepoName), mlog.Int("pr", pr.Number), mlog.String("Fullname", pr.FullName))
+		return
+	}
+	// if builds are 0 means no build ran for master and most problaby this is not setup, so skipping.
+	if len(builds) == 0 {
+		fmt.Println("looks like there is not circleci setup or master never ran. Skipping")
```
hanzei:
```suggestion
		mlog.Debug("looks like there is not circleci setup or master never ran. Skipping")
```

server/config.go
```diff
@@ -44,6 +44,7 @@ type ServerConfig struct {
 	Username                    string
 	AutoAssignerTeam            string
 	AutoAssignerTeamID          int64
+	CircleCiToken               string
```
hanzei: Nit:
```suggestion
	CircleCIToken               string
```

cpanato: done

server/circleci.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/mattermost/mattermost-mattermod/model"
+	"github.com/mattermost/mattermost-server/mlog"
+
+	"github.com/cpanato/go-circleci"
+)
+
+func (s *Server) triggerCircleCiIfNeeded(pr *model.PullRequest) {
+	client := &circleci.Client{Token: s.Config.CircleCIToken}
+
+	if strings.Contains(pr.FullName, "mattermost/") {
```
lieut-data: `strings.Contains` seems like it might match things that aren't from the `mattermost` repo, e.g. `github.com/lieut-data/mattermost`?

cpanato: Agree, thanks for catching that.
the `Fullname` is something like this `lieu-data/reepo1` there is no `/` in the end. but I will update this to another way to check

cpanato: @lieut-data PTAL

cpanato:
@lieut-data PTAL

lieut-data:
@cpanato, feel free to hit the re-request review button: it's the best way to surface a list of PRs for me to review: 
image

cpanato:
ok @lieut-data, i'm old fashion like the ptal :p

lieut-data:
Haha, either works! My notification is pretty backlogged these days, but I do try to prioritize review requests :)

PR #59, opened 2019-10-23
add merge pr block status when set a do not merge label

cpanato:
When we add (for example) the `do not merge` label this will add a status check pending making the PR not mergeable. 
When removing the label it will set the status to success and will be able to merge the PR

This is just  to visually block the PR when we define a label (for example `do not merge`)

i think this might be useful


lieut-data:
This is a brilliant idea!

cpanato:
@levb tested locally in a sample repo that i have :)

go.mod
```diff
@@ -29,4 +29,4 @@ require (
 	k8s.io/api v0.0.0-20190814101207-0772a1bdf941 // indirect
 )
 
-replace git.apache.org/thrift.git => github.com/apache/thrift v0.12.0
+// replace git.apache.org/thrift.git => github.com/apache/thrift v0.12.0
```
hanzei: Why did you remove this? I thought `git.apache.org/thrift.git` is still down

server/block_pr_merge.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+
+	client := NewGithubClient(s.Config.GithubAccessToken)
+	mlog.Info("will block PR merge status", mlog.Int("pr", pr.Number), mlog.String("repo", pr.RepoName))
+	_, _, errStatus := client.Repositories.CreateStatus(context.Background(), pr.RepoOwner, pr.RepoName, pr.Sha, mergeStatus)
+	if errStatus != nil {
+		mlog.Error("Unable to create the github status for for PR", mlog.Int("pr", pr.Number), mlog.Err(errStatus))
+		return
+	}
+
```
hanzei: Nit: Remove this line

server/pull_request.go
```diff
@@ -75,6 +81,9 @@ func (s *Server) handlePullRequestEvent(event *PullRequestEvent) {
 				s.handleUpdateSpinWick(pr, false)
 			}
 		}
+		if s.isBlockPRMergeInLabels(pr.Labels) {
```
hanzei: Why do we need to check for this here? Isn't checking on label switches enough?

server/pull_request.go
```diff
@@ -413,3 +422,16 @@ func (s *Server) CleanOutdatedPRs() {
 	}
 	mlog.Info("Finished update the outdated prs in the mattermod database....")
 }
+
+func (s *Server) isBlockPRMerge(label string) bool {
+	return label == s.Config.BlockPRMergeLabel
```
hanzei: Maybe we check if `s.Config.BlockPRMergeLabel` is a prefix of `label`. This way we can set `s.Config.BlockPRMergeLabel` to `Do Not Merge` and don't have to configure the individual labels.

server/pull_request.go
cpanato: yes, but this is just to update the check to the latest commit hash. i can remove @hanzei

hanzei: Makes sense :+1:

PR #63, opened 2019-11-12
add update branch command

cpanato:
This adds the ability to update the PR with the latest upstream.
this is handy when the Core Members need to merge a PR and that needs an update.

the command is `/update-branch` and will try to update the branch if it fails will post a message and you will need to do manually

### DIFF
--- a/server/update_branch.go
+++ b/server/update_branch.go
@@ -1,0 +1,48 @@
+package server
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"strings"
+
+	"github.com/mattermost/mattermost-server/mlog"
+
+	"github.com/google/go-github/v28/github"
+)
+
+func (s *Server) handleUpdateBranch(eventIssueComment IssueComment) {
+	client := NewGithubClient(s.Config.GithubAccessToken)
+	prGitHub, _, err := client.PullRequests.Get(context.Background(), *eventIssueComment.Repository.Owner.Login, *eventIssueComment.Repository.Name, *eventIssueComment.Issue.Number)
+	if err != nil {
+		mlog.Error("Error getting the latest PR information from github", mlog.Err(err))
+		return
+	}
+	pr, err := s.GetPullRequestFromGithub(prGitHub)
+	if err != nil {
+		mlog.Error("Error Updating the PR in the DB", mlog.Err(err))
+		return
+	}
+
+	userComment := *eventIssueComment.Comment.User
+	if !s.checkUserPermission(userComment.GetLogin(), pr.RepoOwner) {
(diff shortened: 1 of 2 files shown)


server/update_branch.go
```diff
@@ -0,0 +1,45 @@
...
+		return
+	}
+	pr, err := s.GetPullRequestFromGithub(prGitHub)
+	if err != nil {
+		mlog.Error("Error Updating the PR in the DB", mlog.Err(err))
+		return
+	}
+
+	userComment := *eventIssueComment.Comment.User
+	if !s.checkUserPermission(userComment.GetLogin(), pr.RepoOwner) {
```
hanzei: If the PR is made from a fork this will fail, even though `userComment` might have access. Can we check `pr.MaintainerCanModify` (https://godoc.org/github.com/google/go-github/github#PullRequest) if the check fails?

server/update_branch.go
```diff
@@ -0,0 +1,45 @@
...
+		s.sendGitHubComment(pr.RepoOwner, pr.RepoName, pr.Number, "Looks like you dont have permissions to trigger this command.\n Only available for Org members")
+		return
+	}
+
+	opt := &github.PullReqestBranchUpdateOptions{
+		ExpectedHeadSHA: github.String(pr.Sha),
+	}
+
+	_, resp, err := client.PullRequests.UpdateBranch(context.Background(), pr.RepoOwner, pr.RepoName, pr.Number, opt)
```
hanzei: Does this work for forks?

server/update_branch.go
```diff
@@ -0,0 +1,45 @@
...
+	if resp.StatusCode != 202 {
```
hanzei: Nit:
```suggestion
	if resp.StatusCode != http.StatusAccepted {
```

server/update_branch.go
```diff
@@ -0,0 +1,45 @@
...
+		s.sendGitHubComment(pr.RepoOwner, pr.RepoName, pr.Number, "Error trying to update the PR.\nPlease do it manually.")
+		return
+	}
+	if err != nil {
+		if !strings.Contains("job scheduled on GitHub side; try again later", err.Error()) {
+			s.sendGitHubComment(pr.RepoOwner, pr.RepoName, pr.Number, "Error trying to update the PR.\nPlease do it manually.")
```
hanzei: Can we also post the error please?

server/update_branch.go
cpanato: github api does not say that dont work for forks, i think it will

server/update_branch.go
cpanato: this is the same for other command and i've test that with a user that is not part of the org

hanzei: Sorry, I was not clear on this:
If the PR is created from a fork, we should check `pr.MaintainerCanModify` and post a message that the PR creator should allow this.

PR #66, opened 2019-11-29
IS-234: Trigger circleci for PRs on "Build" Label added

metanerd:
This builds apps for mobile for PRs in a separate prefixed branch, then cleans up the branch after it successfully finished in the next tick, when the branch has the specific prefix. 
This is a workaround for circleci not being able to trigger parameterized workflows for PRs.

https://mattermost.atlassian.net/browse/IS-234


server/github.go
```diff
@@ -168,3 +171,54 @@ func (s *Server) checkUserPermission(user, repoOwner string) bool {
...
 }
+
+func (s *Server) createRefWithPrefixFromPr(pr *model.PullRequest, prefix string) {
+	client := NewGithubClient(s.Config.GithubAccessToken)
+	_, _, err := client.Git.CreateRef(
+		context.Background(),
+		pr.RepoOwner,
+		pr.RepoName,
+		&github.Reference{
+			Ref: github.String("refs/heads/" + prefix + strconv.Itoa(pr.Number)),
```
cpanato: `/ref/heads/PRNUMBER`is the correct reference? i thought is like `refs/pull/PRNUMBER/head`

server/github.go
```diff
@@ -168,3 +171,54 @@ func (s *Server) checkUserPermission(user, repoOwner string) bool {
...
+	client := NewGithubClient(s.Config.GithubAccessToken)
+
+	branches, _, err := client.Repositories.ListBranches(context.Background(), repoOwner, repoName, nil)
+	if err != nil {
+		mlog.Error("Error listing branches", mlog.Err(err))
+	}
+
+	regexBranchPrefix := regexp.MustCompile(`^` + prefix + `[0-9]+$`)
+
+	for _, branch := range branches {
```
cpanato: this will delete other branches that might be running and needed. can we delete only the requested one?

server/build_mobile_apps.go
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) 2017-present Mattermost, Inc. All Rights Reserved.
+// See License.txt for license information.
+
+package server
+
+import (
+	"github.com/mattermost/mattermost-mattermod/model"
+)
+
+func (s *Server) buildMobileApp(pr *model.PullRequest) {
```
cpanato: this will just create the new branch, how we know the pipeline for this new branch passed? and also i think we need to comment in the PR the links to download the built packages

server/github.go
metanerd: According to the api docs (https://developer.github.com/v3/git/refs/#create-a-reference), this is the destination reference, which will be created. It fetches the PR commit via the given pr.Sha

server/build_mobile_apps.go
metanerd: Do you mean, if the PR checks themselves have passed before trying to build the mobile package? I forgot about that. Makes totally sense to add it, good catch. Not yet sure how to get the link to the built packages. Maybe it could also be done in the circleci part of the repo instead and posted into a channel (that was @enahum's workaround suggestion). I will find a solution.

enahum: currently it posts to the QA channel, but if we can construct them and post into the PR that would be AWESOME!

enahum:
@metanerd @cpanato where are we with this one ?

metanerd:
Almost there. Really sorry it takes so long.

server/github.go
metanerd: Fixed this one and will only delete the requested one. Way better now, thank you.

PR #70, opened 2019-12-20
Add GitHub webhook secret validation

bpietraga:
Add support for simple GitHub webhook secret validation.

#### Summary

Perform validation based on one shared (between repositories) webhook secret.

This is simple approach, as the repositories can have multiple webhook secrets. In this PR there is one predefined secret with config key `GitHubWebhookSecret `

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/13347


config/config-mattermost.default.json
```diff
@@ -79,4 +80,4 @@
     "FileFormat": "",
     "FileLocation": ""
   }
-}
\ No newline at end of file
+}
```
cpanato: new line here :)

server/validate_signature.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"crypto/hmac"
+	"crypto/sha1"
+	"encoding/hex"
+	"errors"
+	"fmt"
+)
+
+func validateSignature(recievedHash []string, bodyBuffer []byte, key string) error {
```
metanerd: Thank you for your work on this! I think there is a typo in recieve. Maybe a test would be nice for this validate function, what do you think?

bpietraga: Fully agree!

bpietraga: Fully agree this one, thanks for catching typo!

bpietraga: I've make func public, added comment and tests :+1:

config/config-mattermost.default.json
bpietraga: Removed :)

bpietraga:
**Note to reviewers.** This is breaking change, as the code enforces the usage of Secret! The validation fails when the `GitHubWebhookSecret ` is not provided!

If there is a need to keep the code backwards compatible with webhooks setups where the secret is not used like:
`"GitHubWebhookSecret": ""` 

Then the execution can be conditional only if the config holds non empty value of `GitHubWebhookSecret`. 
So it would be solved with simple `if` statement if needed.

bpietraga:
> 🎉 I received the error without the secret on the ping event, after adding it and redelivering, the ping event succeeded! Thank you so much, @bpietraga !

Great to hear! Thanks for suggestions @metanerd  🎉

PR #76, opened 2020-01-31
Add a check for when integration files are being modified.

metanerd:
When integration files are being touched, e.g. mattermost-server file interfacing with mmctl, make the author and reviewers of the PR aware to check, if the integration still works.


hanzei:
@metanerd Sorry for the late review. I quite like to refactoring to help that allows adding tests.

But I'm wondering what issue this PR tries to address. What makes `model/client4.go` so special in comparison to other files in `model` that can also cause a breaking change?

metanerd:
@hanzei maybe it would be good to add all the model's into it. Maybe I just wasn't aware enough.

hanzei:
I'm wondering if there a more elegant way to check if there are breaking changes. The current way seams a bit to spamy (see https://github.com/mattermost/mattermost-server/pull/14023) and doesn't catch every breaking change, as changes in `app` can also change the behaviour.

metanerd:
This is the blacklist check, not the integration files check. The blacklist check should actually spam. The blacklist check can also be easily remedied by creating the branch upstream and not in a fork.

hanzei:
Got it, thanks for the hint. :+1:

hanzei:
Got it, thanks for the hint. :+1:

PR #77, opened 2020-02-07
Instead of listing all docker tags for a repo, check if the manifest is available for that tag

jwilander:
Not sure why the tags request was failing (happened locally for me too) but this is a better way to do it anyways. Tested locally.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,6 +16,7 @@ require (
 	github.com/gorilla/mux v1.7.3
 	github.com/heroku/docker-registry-client v0.0.0-20190909225348-afc9e1acc3d5
 	github.com/mattermost/mattermost-cloud v0.7.0
+	github.com/mattermost/mattermost-server v5.11.1+incompatible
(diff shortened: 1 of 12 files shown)


jwilander:
Build failure looks unrelated

cpanato:
we need to make the build Pass, otherwise cannot deploy

jwilander:
Read my summary @cpanato :P it was a hacky test but it was a test

go.mod
```diff
@@ -16,6 +16,7 @@ require (
 	github.com/gorilla/mux v1.7.3
 	github.com/heroku/docker-registry-client v0.0.0-20190909225348-afc9e1acc3d5
 	github.com/mattermost/mattermost-cloud v0.7.0
+	github.com/mattermost/mattermost-server v5.11.1+incompatible
```
hanzei: @jwilander Why is the import of this version needed?

jwilander: I didn't pick that version specifically. I just ran `go mod vendor` to fix the lint issues so I could get this merged.

PR #83, opened 2020-02-14
IS-234: Use a different approach for building mobile apps.

metanerd:
1. Check first for running builds
1. No running builds found, check for finished builds
1. Finished builds should have the expected amount of jobs and expected job names
1. Finished artifacts for each expected job should have the expected amount of artifacts

### DIFF
--- a/server/build_mobile_apps_test.go
+++ b/server/build_mobile_apps_test.go
@@ -5,26 +5,26 @@ import (
 	"testing"
 )
 
-func TestFindExpectedArtifacts(t *testing.T) {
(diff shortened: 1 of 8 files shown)


server/build_mobile_apps_test.go
```diff
@@ -0,0 +1,48 @@
+package server
+
+import (
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestFindExpectedArtifacts(t *testing.T) {
```
jaydeland: Awesome to have some build tests

Issue #73, opened 2020-01-17
Allow use of update-branch by PR author.


lieut-data:
Currently, the `update-branch` command is restricted to members of the corresponding repository, but this command is very useful for community authors as well.

Investigate the security ramifications, and then relax this check to allow the author of the PR to effect this command as well.

PR #92, opened 2020-04-03
Allow submitter of a PR to user /update-branch

hanzei:
#### Summary
Head up that I didn't test this locally as I don't know how.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-mattermod/issues/73

### DIFF
--- a/server/update_branch.go
+++ b/server/update_branch.go
@@ -24,10 +24,13 @@ func (s *Server) handleUpdateBranch(eventIssueComment IssueComment) {
 		return
 	}
 
-	userComment := *eventIssueComment.Comment.User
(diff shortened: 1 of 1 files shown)


hanzei:
I found it hard to add tests given that `checkUserPermission` uses `github.Client` to fetch the org membership. The proper way to test with would to mock `github.Client` and then unit test `checkUserPermission`. But that seams like a change that should be done holistically.

hanzei:
@metanerd What do think about my thoughts? Do you have ideas on how to refactor the code to make it testable?

metanerd:
> @metanerd What do think about my thoughts? Do you have ideas on how to refactor the code to make it testable?

Agreed.

PR #107, opened 2020-04-28
Fix mobile build review process.

metanerd:
#### Summary
https://developer.github.com/v3/repos/statuses/#get-the-combined-status-for-a-specific-ref
Since we changed the review process to be pending, this check needs to be adapted.

### DIFF
--- a/server/github.go
+++ b/server/github.go
@@ -258,7 +258,7 @@ func (s *Server) areChecksSuccessfulForPr(pr *model.PullRequest, org string) (bo
 		return false, err
 	}
 	mlog.Debug("Retrieved status for pr", mlog.String("status", cStatus.GetState()), mlog.Int("prNumber", pr.Number), mlog.String("prSha", pr.Sha))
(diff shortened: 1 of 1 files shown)


server/github.go
```diff
@@ -258,7 +258,7 @@ func (s *Server) areChecksSuccessfulForPr(pr *model.PullRequest, org string) (bo
 		return false, err
 	}
 	mlog.Debug("Retrieved status for pr", mlog.String("status", cStatus.GetState()), mlog.Int("prNumber", pr.Number), mlog.String("prSha", pr.Sha))
-	if cStatus.GetState() == "success" || cStatus.GetState() == "" {
+	if cStatus.GetState() == "success" || cStatus.GetState() == "pending" || cStatus.GetState() == "" {
```
enahum, line 261: so pending allows us to build the app while reviews are still to be done ?

metanerd: yes :)

PR #110, opened 2020-04-29
IS-255: Poll build link

metanerd:
#### Summary
Add logic to poll the build link for the triggered enterprise tests.

### DIFF
--- a/server/circleci.go
+++ b/server/circleci.go
@@ -83,11 +83,11 @@ type PipelineTriggeredResponse struct {
 	CreatedAt time.Time `json:"created_at"`
 }
 
-func (s *Server) triggerEnterprisePipeline(ctx context.Context, pr *model.PullRequest, eeBranch string, triggerBranch string) (string, error) {
+func (s *Server) triggerEnterprisePipeline(ctx context.Context, pr *model.PullRequest, eeBranch string, triggerBranch string) error {
 	body := strings.NewReader(`branch=` + eeBranch + `&parameters[external_branch]=` + triggerBranch + `&parameters[external_sha]=` + pr.Sha + `&parameters[external_pr]=` + strconv.Itoa(pr.Number))
 	req, err := http.NewRequest("POST", "https://circleci.com/api/v2/project/gh/"+s.Config.Org+"/"+s.Config.EnterpriseReponame+"/pipeline", body)
 	if err != nil {
-		return "", err
+		return err
 	}
 	req.SetBasicAuth(os.ExpandEnv(s.Config.CircleCIToken), "")
 	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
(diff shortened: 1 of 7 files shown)


server/circleci.go
```diff
@@ -74,22 +76,117 @@ func (s *Server) triggerCircleCiIfNeeded(pr *model.PullRequest) {
...
+		return "", err
 	}
 	req.SetBasicAuth(os.ExpandEnv(s.Config.CircleCIToken), "")
 	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
 
-	resp, err := http.DefaultClient.Do(req)
+	client := http.Client{
+		Timeout: 20 * time.Second,
+	}
+	resp, err := client.Do(req)
```
agnivade, line 98: We can set the timeout in the request itself, using `http.NewRequestWithContext` or the `req.WithContext` method. Initializing a new http.Client every time has a cost since a client is a fat object maintaining persistent TCP connections with many hosts and reuses them.

server/circleci.go
```diff
@@ -74,22 +76,117 @@ func (s *Server) triggerCircleCiIfNeeded(pr *model.PullRequest) {
...
 	mlog.Debug("EE triggered", mlog.Int("pr", pr.Number), mlog.String("sha", pr.Sha), mlog.String("triggerRef", triggerBranch), mlog.String("eeBranch", eeBranch))
 	if err != nil {
-		return err
+		return "", err
+	}
+	b, err := ioutil.ReadAll(resp.Body)
```
agnivade, line 103: The initial error check is missed here.

But this entire thing can be condensed to 

`err := json.NewDecoder(resp.Body).Decode(&triggeredR)`. No need to read and unmarshal. And the resp.Body.Close() can be in a defer clause, as we would want to close the body even if there was an error in reading.

server/circleci.go
```diff
@@ -74,22 +76,117 @@ func (s *Server) triggerCircleCiIfNeeded(pr *model.PullRequest) {
...
+
+func (s *Server) waitForWorkflowId(ctx context.Context, id string, workflowName string) (string, error) {
+	ticker := time.NewTicker(10 * time.Second)
+	for {
+		select {
+		case <-ctx.Done():
+			ticker.Stop()
+			return "", errors.New("timed out trying to fetch workflow")
+		case <-ticker.C:
+			req, err := http.NewRequest("GET", "https://circleci.com/api/v2/pipeline/"+id+"/workflow", nil)
```
agnivade, line 149: Let's add a timeout here too.

server/circleci.go
```diff
@@ -74,22 +76,117 @@ func (s *Server) triggerCircleCiIfNeeded(pr *model.PullRequest) {
...
+			if err != nil {
+				return "", err
+			}
+			req.SetBasicAuth(os.ExpandEnv(s.Config.CircleCIToken), "")
+
+			resp, err := http.DefaultClient.Do(req)
+			if err != nil {
+				return "", err
+			}
+			b, err := ioutil.ReadAll(resp.Body)
```
agnivade, line 159: Same as above. Apply json.NewDecoder + Decode combo. Close the body on defer.

server/circleci.go
```diff
@@ -74,22 +76,117 @@ func (s *Server) triggerCircleCiIfNeeded(pr *model.PullRequest) {
...
+				if pip.Name == workflowName {
+					workflowId = pip.WorkflowId
+				}
+			}
+
+			if workflowId == "" {
+				return "", fmt.Errorf("unable to find workflow, %v", err)
+			}
+
+			ticker.Stop()
```
agnivade, line 186: The issue is that ticker won't be stopped in the other return paths. We should declare this in a defer call right after `NewTicker`.

server/github.go
```diff
@@ -263,3 +265,44 @@ func (s *Server) areChecksSuccessfulForPr(pr *model.PullRequest, org string) (bo
...
+				if *status.Context == statusContext && *status.State == statusState {
+					hasStatus = true
+				}
+			}
+
+			if !hasStatus {
+				continue
+			}
+
+			ticker.Stop()
```
agnivade, line 304: Same as above. Move this to a defer right after NewTicker

server/github.go
```diff
@@ -263,3 +265,44 @@ func (s *Server) areChecksSuccessfulForPr(pr *model.PullRequest, org string) (bo
 	}
 	return false, nil
 }
+
+func (s *Server) createRepoStatus(pr *model.PullRequest, status *github.RepoStatus) {
+	client := NewGithubClient(s.Config.GithubAccessToken)
+	_, _, err := client.Repositories.CreateStatus(context.Background(), pr.RepoOwner, pr.RepoName, pr.Sha, status)
```
agnivade, line 271: This creates a new context again, whereas the parent context can be trickled down from `updateBuildStatus`.

metanerd:
Thank you for your review @agnivade ! This is very helpful to know for the overall code. I will address these issues in a followup PR.

agnivade:
Amazing :)

server/github.go
metanerd: Good to know. This is all over mattermod honestly.

PR #111, opened 2020-04-30
Overall code improvements. 

metanerd:
#### Summary
Code improvements. See #110

### DIFF
--- a/server/circleci.go
+++ b/server/circleci.go
@@ -10,7 +10,6 @@ import (
 	"github.com/mattermost/mattermost-mattermod/model"
 	"github.com/mattermost/mattermost-server/v5/mlog"
 	"github.com/pkg/errors"
-	"io/ioutil"
 	"net/http"
 	"os"
 	"strconv"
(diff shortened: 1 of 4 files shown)


server/circleci.go
```diff
@@ -76,51 +75,63 @@ func (s *Server) triggerCircleCiIfNeeded(pr *model.PullRequest) {
...
-	}
-	resp, err := client.Do(req)
+	resp, err := circleciClient.Do(req)
 	mlog.Debug("EE triggered", mlog.Int("pr", pr.Number), mlog.String("sha", pr.Sha), mlog.String("triggerRef", triggerBranch), mlog.String("eeBranch", eeBranch))
 	if err != nil {
-		return "", err
+		return nil, err
 	}
-	b, err := ioutil.ReadAll(resp.Body)
 	err = resp.Body.Close()
```
agnivade, line 124: We should not close the body before reading it. This should be done in a defer block.

server/circleci.go
```diff
@@ -76,51 +75,63 @@ func (s *Server) triggerCircleCiIfNeeded(pr *model.PullRequest) {
 	mlog.Info("Triggered circleci", mlog.String("repo", pr.RepoName), mlog.Int("pr", pr.Number), mlog.String("fullname", pr.FullName))
 }
 
+func (s *Server) requestEETriggering(ctx context.Context, pr *model.PullRequest, eeBranch string, triggerBranch string) error {
+	circleciClient := http.DefaultClient
```
agnivade, line 79: If we are passing the default client, then there's no need to pass it really. We can just use that directly in that method.

server/circleci.go
```diff
@@ -139,42 +150,37 @@ type PipelineWorkflowResponse struct {
...
 			}
 			req.SetBasicAuth(os.ExpandEnv(s.Config.CircleCIToken), "")
 
-			resp, err := http.DefaultClient.Do(req)
+			resp, err := circleciClient.Do(req)
 			if err != nil {
 				return "", err
 			}
-			b, err := ioutil.ReadAll(resp.Body)
 			err = resp.Body.Close()
```
agnivade, line 171: Same as above.

server/pull_request.go
```diff
@@ -30,8 +30,10 @@ func (s *Server) handlePullRequestEvent(event *PullRequestEvent) {
 		s.addHacktoberfestLabel(pr)
 
 		if pr.RepoName == s.Config.EnterpriseTriggerReponame {
-			s.createEnterpriseTestsPendingStatus(pr)
-			go s.triggerEETestsforOrgMembers(pr)
+			ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
```
agnivade, line 33: Might be useful to refactor the `5 * time.Minute` as a constant somewhere else since this is being used 4 times.

metanerd:
> Great improvements ! I just have some small comments to improve a bit more.
> 
> If we really want it to take it to the next level, then the next step should be to use a job queue to handle all the goroutines. Otherwise right now, we are creating goroutines in an unbounded manner which is not a good thing. But yes, this is totally a nice-to-have.

I would like to implement a job queue. Right now I see a more pressing need on the other end of the code which I would like to address first.

server/circleci.go
```diff
@@ -141,40 +147,33 @@ type PipelineWorkflowResponse struct {
...
-			}
-
 			if resp.StatusCode != http.StatusOK {
 				continue
 			}
-
-			triggeredR := PipelineWorkflowResponse{}
-			err = json.Unmarshal(b, &triggeredR)
+			r := PipelineWorkflowResponse{}
+			defer resp.Body.Close()
```
agnivade, line 169: This is a bug. The defer call should be made immediately after the err check. Right now, the body won't be closed if we get a 500 response.

server/pull_request.go
metanerd: I think, I should fine tune this not to be such a constant value instead. Would it add a lot of value to be a constant?

agnivade: Do you foresee changing the timeout value dynamically during program execution ? That would be very surprising. We usually always default to a constant if we don't change a value.

PR #117, opened 2020-04-30
Cleanup labels when PR gets closed

hanzei:
#### Summary

Cleanup labels when a PR gets closed that don't make sense on a closed PR like `Setup Cloud Test Server` or `Do Not Merge`.

I stole the mock testing approach from https://github.com/mattermost/matterbuild/pull/28. All credit goes to @ali-farooq0. This will allow much easier testing in the future.


server/pull_request.go
```diff
@@ -457,6 +469,44 @@ func (s *Server) CleanOutdatedPRs() {
...
+	var wg sync.WaitGroup
+
+	for _, l := range labels {
+		for _, labelToRemove := range labelsToRemove {
+			if l.GetName() == labelToRemove {
+				wg.Add(1)
+				go func(label string) {
+					defer wg.Done()
+					s.removeLabel(client, pr.RepoOwner, pr.RepoName, pr.Number, label)
+				}(labelToRemove)
```
hanzei, line 502: I did go with a separate go routine here for better performance.

server/pull_request.go
```diff
@@ -457,6 +469,44 @@ func (s *Server) CleanOutdatedPRs() {
...
+			}
+		}
+	}
+
+	wg.Wait()
```
hanzei, line 507: Blocking here to allow easier tesing.

server/pull_request.go
```diff
@@ -457,6 +469,44 @@ func (s *Server) CleanOutdatedPRs() {
 	mlog.Info("Finished update the outdated prs in the mattermod database....")
 }
 
+func (s *Server) CleanUpLabels(client *GithubClient, pr *model.PullRequest) {
```
hanzei, line 472: I thought about returning an `error`, but that would mean that the caller can not longer fire-and-forget. Maybe it makes sense to start migrating methods to return proper errors.

server/pull_request.go
```diff
@@ -457,6 +468,44 @@ func (s *Server) CleanOutdatedPRs() {
...
+		"Work In Progress",
+	}
+
+	labels, _, err := client.Issues.ListLabelsByIssue(context.Background(), pr.RepoOwner, pr.RepoName, pr.Number, nil)
+	if err != nil {
+		mlog.Error("Error listing the labels for closed PR", mlog.Err(err))
+		return
+	}
+
+	var wg sync.WaitGroup
```
ali-farooq0, line 492: I would recommend using `var g errgroup.Group` here instead of this as that will help with the easier transition to when the go-routine functions start returning errors. It'll just set it up nicely for that. Otherwise someone will have to add the error handling logic here, which `errorgroup.Group` already does inside.

https://www.oreilly.com/content/run-strikingly-fast-parallel-file-searches-in-go-with-sync-errgroup/

Code looks much more clean. Just a suggestion :)

server/pull_request_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+	"github.com/golang/mock/gomock"
+	"github.com/google/go-github/v28/github"
+
+	"github.com/mattermost/mattermost-mattermod/model"
+	"github.com/mattermost/mattermost-mattermod/server"
+	"github.com/mattermost/mattermost-mattermod/server/mocks"
+)
+
+func TestCleanUpLabels(t *testing.T) {
```
ali-farooq0, line 15: nicely done 👍

server/pull_request.go
ali-farooq0: you can still return the `error` here (and any new functions that you deem necessary) and just have the caller ignore it, that way when the caller is ready to transition to start checking, it can. Just one less thing to worry about at that point.

hanzei: @ali-farooq0 How would the caller properly check the error? Something like 
```golang
go func() {
	err := CleanUpLabels(...)
	if err != nil {
		Log.Stuff(...)
	}
}()
```

I'm unsure how much I would like that pattern.

ali-farooq0: Fair point. I only recommended returning the `error` if the idea was to handle all the errors at the top level at some point (and migrate other functions to do the same). But doing that would mean that you'd have to do what you highlighted above, which is not the cleanest. Usually you return an error when you want the caller to stop the operation when it receives the error, but in this case it'll be handled through the logs it seems.

server/pull_request.go
ali-farooq0: seeing our discussion above, no major advantage of using `errorgroup` then since we don't plan to return any errors in the near future.

metanerd:
This PR is great! Thank you so much @hanzei for your effort in stealing @ali-farooq0 's approach!

PR #123, opened 2020-05-01
Add CLA exclusion list of users

khos2ow:
#### Summary

Add a new config `CLAExclusionsList` to explicitly exclude users from signing CLA. For example we recently enabled `dependabot` and since it's a known bot it doesn't need to sign the CLA.


server/cla.go
```diff
@@ -121,3 +130,15 @@ func checkCLAComment(comments []*github.IssueComment, username string) (int64, b
 	}
 	return 0, false
 }
+
+func isUserExcluded(exclusions []string, username string) bool {
+	if len(exclusions) == 0 {
+		return false
+	}
```
hanzei, line 137: Is this really needed? If `len(exclusions) == 0` the `for` loop won't run anyway.

server/cla.go
```diff
@@ -121,3 +130,15 @@ func checkCLAComment(comments []*github.IssueComment, username string) (int64, b
 	}
 	return 0, false
 }
+
+func isUserExcluded(exclusions []string, username string) bool {
```
hanzei, line 134: Maybe a more generic name like `contains` would fit the method better, as it's not specific for users.

server/cla.go
```diff
@@ -37,9 +37,18 @@ func (s *Server) checkCLA(pr *model.PullRequest) {
...
-		mlog.Int("pr n", pr.Number))
+	mlog.Info(
+		"Will check the CLA for user",
+		mlog.String("user", username),
+		mlog.String("repo", pr.RepoOwner),
+		mlog.String("reponame", pr.RepoName),
+		mlog.Int("pr n", pr.Number),
+	)
+
+	if excluded := isUserExcluded(s.Config.CLAExclusionsList, username); excluded {
```
hanzei, line 48: You can simplify this to
```suggestion
	if isUserExcluded(s.Config.CLAExclusionsList, username) {
```

server/cla.go
hanzei, line 134: 0/5 Move the method to `utils.go` as it's a generic one.

khos2ow: +1

server/cla.go
khos2ow: Yeah that was some leftover, originally I had it like `exlucded, err := isUserExcluded(...)` but dropped the `error` and forgot about this.

PR #129, opened 2020-05-06
Trigger ee tests with appropriate ee branch or fallback to base branch. 

metanerd:
#### Summary
Triggering the ee tests from a server PR will look for the same named enterprise branch, if this branch exists in the upstream. 
Refactors the the branching logic. 
Removes the jenkins github status auto succeed.

### DIFF
--- a/server/enterprise.go
+++ b/server/enterprise.go
@@ -61,12 +61,11 @@ type EETriggerInfo struct {
 }
 
 func (s *Server) getPRInfo(ctx context.Context, pr *model.PullRequest) (info *EETriggerInfo, err error) {
-	clientGitHub := NewGithubClient(s.Config.GithubAccessToken)
-	pullRequest, _, err := clientGitHub.PullRequests.Get(ctx, s.Config.Org, pr.RepoName, pr.Number)
+	pullRequest, _, err := s.GithubClient.PullRequests.Get(ctx, s.Config.Org, pr.RepoName, pr.Number)
 	if err != nil {
 		return nil, err
 	}
-	baseBranch := pullRequest.Base.GetRef()
+	baseBranch := pullRequest.GetBase().GetRef()
 	isFork := pullRequest.GetHead().GetRepo().GetFork()
 	var serverOwner string
 	if isFork {
@@ -86,7 +85,7 @@ func (s *Server) getPRInfo(ctx context.Context, pr *model.PullRequest) (info *EE
 		eeBranch = baseBranch
 	}
 
-	webappOwner, webappBranch, err := s.getSameBranchInRepo(ctx, pr, s.Config.EnterpriseWebappReponame)
+	webappOwner, webappBranch, err := s.getBranchFromForkOrUpstreamRepo(ctx, pr, s.Config.EnterpriseWebappReponame)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


server/enterprise.go
```diff
@@ -77,68 +78,68 @@ func (s *Server) getPRInfo(ctx context.Context, pr *model.PullRequest) (*EETrigg
...
+		WebappBranch: webappBranch,
+	}
+	return info, nil
 }
 
-func (s *Server) getWebappBranchWithSameName(client *github.Client, ctx context.Context, serverPR *github.PullRequest) (owner string, branch string, err error) {
-	prAuthor := serverPR.GetUser().GetLogin()
-	ref := serverPR.GetHead().GetRef()
-	forkBranch, r, err := client.Repositories.GetBranch(ctx, prAuthor, s.Config.EnterpriseWebappReponame, ref)
+func (s *Server) getBranchWithSameName(ctx context.Context, remote string, repo string, ref string) (branch string, err error) {
```
ethervoid, line 109: I think we should pass the GitHub client here in order to avoid creating a new one every time.

I've seen that we're creating that client with the same parameter always so maybe we could move that to the server struct and use it from there and in case we need a new one could be created with a different token in another place

server/enterprise.go
```diff
@@ -77,68 +78,68 @@ func (s *Server) getPRInfo(ctx context.Context, pr *model.PullRequest) (*EETrigg
...
+	clientGitHub := NewGithubClient(s.Config.GithubAccessToken)
+	ghBranch, r, err := clientGitHub.Repositories.GetBranch(ctx, remote, repo, ref)
 	if err != nil {
 		if r == nil {
-			return "", "", err
+			return "", err
```
ethervoid, line 114: You can do something like because golang uses short circuit evaluation:

```go
if r == nil || r.StatusCode != http.StatusNotFound {
    return "", err
}
return "", nil
```

server/enterprise.go
```diff
@@ -77,68 +78,68 @@ func (s *Server) getPRInfo(ctx context.Context, pr *model.PullRequest) (*EETrigg
...
+		return "", nil // do not err if branch is not found
+	}
+	if ghBranch == nil {
+		return "", fmt.Errorf("unexpected failure case")
+	}
+	return ghBranch.GetName(), nil
+}
 
-		upstreamBranch, r, err := client.Repositories.GetBranch(ctx, s.Config.Org, s.Config.EnterpriseWebappReponame, ref)
+func (s *Server) getSameBranchInRepo(ctx context.Context, serverPR *model.PullRequest, repo string) (owner string, branch string, err error) {
```
ethervoid, line 127: This naming is not very accurate because you're searching for the repo in the user's account, as a fork, or in the upstream in case the fork doesn't exists.

I'll call it something like `findBranchInUserOrUpstreamRepo` or something like that

server/enterprise.go
```diff
@@ -77,68 +78,68 @@ func (s *Server) getPRInfo(ctx context.Context, pr *model.PullRequest) (*EETrigg
...
-				return "", "", err
-			}
-			return s.Config.Org, serverPR.GetBase().GetRef(), nil
+			return "", "", err
 		}
-
-		owner := s.Config.Org
-		mlog.Debug("Found upstream webapp branch", mlog.String("owner", owner), mlog.String("branch", upstreamBranch.GetName()))
-		return owner, upstreamBranch.GetName(), nil
+		if upstreamBranch == forkBranch {
```
ethervoid, line 137: I think this should be removed. At this point, `forkBranch` is empty so we just have to check for empty string

server/enterprise.go
```diff
@@ -173,16 +174,6 @@ func (s *Server) createEnterpriseTestsErrorStatus(ctx context.Context, pr *model
 		"Failed running enterprise tests. @mattermost/core-build-engineers have been notified. Error:  \n```"+err.Error()+"```")
 }
 
-func (s *Server) succeedOutDatedJenkinsStatuses(ctx context.Context, pr *model.PullRequest) {
```
ethervoid, line 176: Why did we remove this? (Just to have a bit of context)

metanerd: The transition phase is over, so it should be good to remove.

server/enterprise.go
metanerd: That's the way to properly handle this! Nice, thank you.  I don't want to touch it everywhere in this PR, but I will slowly move it there and will do it for these instances now.

ethervoid: Agree. Thank you!

hanzei: Maybe that's worth a HW in this repo? Community could help with the migration.

ethervoid: I think that's a really good idea :)

PR #130, opened 2020-05-07
Move gh client to server struct.

metanerd:
#### Summary
Fix so that not a new github client is created before every github request. 

#### Ticket Link

### DIFF
--- a/server/auto_merge.go
+++ b/server/auto_merge.go
@@ -6,7 +6,6 @@ package server
 import (
 	"context"
 	"fmt"
-
 	"github.com/google/go-github/v28/github"
 	"github.com/mattermost/mattermost-mattermod/model"
 	"github.com/mattermost/mattermost-server/v5/mlog"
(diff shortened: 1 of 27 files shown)


server/auto_merge.go
```diff
@@ -6,7 +6,6 @@ package server
 import (
 	"context"
 	"fmt"
-
```
hanzei, line 9: It's idiomatic for go code to group all imports from the `std` lib together. This is also the style that https://pkg.go.dev/golang.org/x/tools/cmd/goimports?tab=doc enforces.

Why drop this style?

server/limit_rate_gh.go
```diff
@@ -12,9 +12,7 @@ import (
 
 func (s *Server) CheckLimitRateAndSleep() {
 	mlog.Info("Checking the rate limit on Github and will sleep if need...")
-
```
hanzei, line 15: Nit: Why is this line getting dropped?

hanzei:
Nice cleanup :+1:

server/limit_rate_gh.go
metanerd: The line didn't make the code any more readable in my opinion.

server/auto_merge.go
metanerd: I didn't know about this tool. Makes sense to add it to the CI. I added it.

hanzei: I don't have any objections on using `goimports`. Other Mattermost project using `golangci-lint` as a metalinter. Hence, it makes sense to also integrate `golangci-lint` into this repo, which can also run `goimports`.

PR #135, opened 2020-05-26
Different approach to wait for rate limit. 

metanerd:
#### Summary

This is a different approach to wait until the rate limit is above the specified threshold again. 

Additionally the behavior for CLA checking changed. 
Before this change, the CLA got checked on every push, not just when PR is opened or reopened. 
Adding to that, before it went through every comment, to see if it already created the NeedsToSignCLA comment. In case the contributor signed it, it would delete the comment then after the contributor signed.
Now the pending CLA status is created once for opening/reopening and also on PR sync. And then the `IsAlreadySigned` method checks if github status for CLA shows already successful, which is more go idiomatic.


server/issue.go
```diff
@@ -107,28 +108,29 @@ func (s *Server) CleanOutdatedIssues() {
...
 
 	for _, issue := range issues {
-		ghIssue, _, errIssue := s.GithubClient.Issues.Get(context.Background(), issue.RepoOwner, issue.RepoName, issue.Number)
-		if errIssue != nil {
-			mlog.Error("Error getting Pull Request", mlog.String("RepoOwner", issue.RepoOwner), mlog.String("RepoName", issue.RepoName), mlog.Int("PRNumber", issue.Number), mlog.Err(errIssue))
-			if _, ok := errIssue.(*github.RateLimitError); ok {
-				mlog.Error("GitHub rate limit reached")
-				s.CheckLimitRateAndSleep()
-			}
+		ghIssue, _, err := s.GithubClient.Issues.Get(context.Background(), issue.RepoOwner, issue.RepoName, issue.Number)
```
hanzei, line 114: From looking at the code it seams like there are no uses of `CleanOutdatedIssues`. Can we just drop it alongside `SqlIssueStore.List()` and `SqlIssueStore.ListOpen()`?

server/github.go
```diff
@@ -166,6 +166,10 @@ func (s *Server) getMembers(ctx context.Context) (orgMembers []string, err error
 	var allUsers []*github.User
 	for {
 		users, r, err := s.GithubClient.Organizations.ListMembers(ctx, s.Config.Org, opts)
+		if _, ok := err.(*github.RateLimitError); ok {
+			s.sleepUntilRateLimitAboveTokenReserve()
```
hanzei, line 170: I'm 4/5 that is just buries the root issue. Roughly 4 requests are made there. They should always works.

server/pull_request.go
```diff
@@ -426,26 +426,26 @@ func (s *Server) CleanOutdatedPRs() {
...
 
 	for _, pr := range prs {
-		pull, _, errPull := s.GithubClient.PullRequests.Get(context.Background(), pr.RepoOwner, pr.RepoName, pr.Number)
-		if errPull != nil {
-			mlog.Error("Error getting Pull Request", mlog.String("RepoOwner", pr.RepoOwner), mlog.String("RepoName", pr.RepoName), mlog.Int("PRNumber", pr.Number), mlog.Err(errPull))
-			if _, ok := errPull.(*github.RateLimitError); ok {
-				mlog.Error("GitHub rate limit reached")
-				s.CheckLimitRateAndSleep()
-			}
+		pull, _, err := s.GithubClient.PullRequests.Get(context.Background(), pr.RepoOwner, pr.RepoName, pr.Number)
```
hanzei, line 432: Instead of getting all individual PRs, the code should bulk fetch them and calculate the diff locally. The is of cause more logic but it greatly reduces the number of request. The same comment applies to `CheckPRActivity`.

server/github.go
metanerd: The rate limit api call isn't counted towards the rate limit.

server/pull_request.go
metanerd: I want to address the underlying issue next. But unfortunately this has to do for now.

hanzei: Sure, but the code changes doesn't change the number of API call that mattermod does. It only smooths them over a longer time period. Given that the rate limit failure happen all day I don't think this approach helps much.

server/cla.go
```diff
@@ -16,7 +17,7 @@ import (
 )
 
 func (s *Server) handleCheckCLA(eventIssueComment IssueComment) {
-	prGitHub, _, err := s.GithubClient.PullRequests.Get(context.Background(), *eventIssueComment.Repository.Owner.Login, *eventIssueComment.Repository.Name, *eventIssueComment.Issue.Number)
+	prGitHub, _, err := s.GithubClient.PullRequests.Get(context.TODO(), *eventIssueComment.Repository.Owner.Login, *eventIssueComment.Repository.Name, *eventIssueComment.Issue.Number)
```
hanzei, line 20: 0/5 Why `context.TODO()`?

server/cla.go
```diff
@@ -40,6 +41,10 @@ func (s *Server) checkCLA(pr *model.PullRequest) {
 		return
 	}
 
+	if s.IsAlreadySigned(pr) {
```
hanzei, line 44: When does this case happen? New commits overwrite the old status, which means this is never true for sync events.

server/cla.go
```diff
@@ -54,80 +59,108 @@ func (s *Server) checkCLA(pr *model.PullRequest) {
 		return
 	}
 
+	body, errCSV := s.getCSV()
+	if errCSV != nil {
+		return
+	}
+
+	if !isNameInCLAList(strings.Split(string(body), "\n"), username) {
+		comments, _, err := s.GithubClient.Issues.ListComments(context.TODO(), pr.RepoOwner, pr.RepoName, pr.Number, nil)
```
hanzei, line 68: I'm aware that this is old code, but let me point out that this doesn't  work if there are a lot of commits because paging is not implemented. Maybe a `100` as `PerPage` is fine for now, if implementing paging is to much.

server/cla.go
```diff
@@ -54,80 +59,108 @@ func (s *Server) checkCLA(pr *model.PullRequest) {
...
+		if err != nil {
+			mlog.Error("pr_error", mlog.Err(err))
```
hanzei, line 70: Nit pick: The error message should make clear what failed.

server/cla.go
```diff
@@ -54,80 +59,108 @@ func (s *Server) checkCLA(pr *model.PullRequest) {
...
+			return
+		}
+		_, found := findNeedsToSignCLAComment(comments, s.Config.Username)
+		if !found {
+			s.sendGitHubComment(pr.RepoOwner, pr.RepoName, pr.Number, strings.Replace(s.Config.NeedsToSignCLAMessage, "USERNAME", "@"+username, 1))
```
hanzei, line 75: Separate go routine?

server/cla.go
```diff
@@ -54,80 +59,108 @@ func (s *Server) checkCLA(pr *model.PullRequest) {
...
+		}
+		status := &github.RepoStatus{
+			State:       github.String(stateError),
+			Description: github.String(fmt.Sprintf("%v needs to sign the CLA", username)),
+			TargetURL:   github.String(s.Config.SignedCLAURL),
+			Context:     github.String(s.Config.CLAGithubStatusContext),
+		}
+		mlog.Debug("will post error on CLA", mlog.String("user", username))
+		_, _, errStatus := s.GithubClient.Repositories.CreateStatus(context.TODO(), pr.RepoOwner, pr.RepoName, pr.Sha, status)
```
hanzei, line 84: Does it make sense to use `createRepoStatus` here?

server/cla.go
```diff
@@ -54,80 +59,108 @@ func (s *Server) checkCLA(pr *model.PullRequest) {
...
+		return
+	}
+}
+
+func (s *Server) getCSV() ([]byte, error) {
 	resp, err := http.Get(s.Config.SignedCLAURL)
 	if err != nil {
 		mlog.Error("Unable to get CLA list", mlog.Err(err))
-		return
+		s.logToMattermost("unable to get CLA google csv file Error: ```" + err.Error() + "```")
```
hanzei, line 110: :100:

server/cla.go
```diff
@@ -54,80 +59,108 @@ func (s *Server) checkCLA(pr *model.PullRequest) {
 		return
 	}
 
+	body, errCSV := s.getCSV()
```
hanzei, line 62: Out of scope of this PR: Cache the csv for one minute

server/cla.go
```diff
@@ -54,80 +59,108 @@ func (s *Server) checkCLA(pr *model.PullRequest) {
...
-					mlog.Error("Unable to remove old Mattermod comment", mlog.Err(err))
-				}
-			}
-			return
+func isNameInCLAList(usersWhoSignedCLA []string, author string) bool {
+	authorLowerCase := strings.ToLower(author)
+	for _, userToTrim := range usersWhoSignedCLA {
+		user := strings.TrimSpace(userToTrim)
+		if strings.Compare(user, author) == 0 || strings.Compare(user, authorLowerCase) == 0 || strings.Compare(strings.ToLower(user), authorLowerCase) == 0 {
+			return true
```
hanzei, line 138: Nit: this can be simplified 

```suggestion
	for _, userToTrim := range usersWhoSignedCLA {
		user := strings.ToLower(strings.TrimSpace(userToTrim))
		if strings.Compare(user, author) == 0 {
			return true
```

server/cla.go
```diff
@@ -54,80 +59,108 @@ func (s *Server) checkCLA(pr *model.PullRequest) {
...
+		State:       github.String(statePending),
+		Description: github.String("Checking if " + pr.Username + " signed CLA"),
+		TargetURL:   github.String(s.Config.SignedCLAURL),
+		Context:     github.String(s.Config.CLAGithubStatusContext),
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), requestTimeoutGithub)
+	defer cancel()
+	err := s.createRepoStatus(ctx, pr, status)
+	if err != nil {
+		s.logToMattermost("failed to create status for PR: " + strconv.Itoa(pr.Number) + " Context: " + s.Config.CLAGithubStatusContext + " Error: ```" + err.Error() + "```")
```
hanzei, line 164: Drive by idea: Maybe all error of a certain log level should to to Mattermost, e.g. `Error`

server/cla_integration_test.go
```diff
@@ -0,0 +1,120 @@
+package server_test
```
hanzei, line 1: :100:

server/cla_unit_test.go
```diff
@@ -4,6 +4,7 @@ import (
 	"testing"
 
 	"github.com/google/go-github/v31/github"
+
```
hanzei, line 7: Nit: Why is this empty line needed?

server/config.go
```diff
@@ -102,7 +102,9 @@ type Config struct {
 	StartLoadtestTag     string
 	StartLoadtestMessage string
 
-	CLAExclusionsList     []string
+	CLAExclusionsList      []string
+	CLAGithubStatusContext string
```
hanzei, line 106: 0/5 if this needs to be configurable. Maybe using a static string is fine

server/enterprise.go
```diff
@@ -28,7 +29,7 @@ func (s *Server) triggerEnterpriseTests(pr *model.PullRequest) {
 		return
 	}
 
-	isBaseBranchReleaseBranch, err := regexp.MatchString(`$release-*`, triggerInfo.BaseBranch)
+	isBaseBranchReleaseBranch, err := regexp.MatchString(`^release-*`, triggerInfo.BaseBranch)
```
hanzei, line 32: Woops

server/enterprise.go
```diff
@@ -134,12 +135,15 @@ func (s *Server) getBranchFromForkOrUpstreamRepo(ctx context.Context, serverPR *
...
 	enterpriseStatus := &github.RepoStatus{
-		State:       github.String("pending"),
+		State:       github.String(statePending),
 		Context:     github.String(s.Config.EnterpriseGithubStatusContext),
 		Description: github.String("TODO as org member: After reviewing please trigger label \"" + s.Config.EnterpriseTriggerLabel + "\""),
 		TargetURL:   github.String(""),
 	}
-	s.createRepoStatus(ctx, pr, enterpriseStatus)
+	err := s.createRepoStatus(ctx, pr, enterpriseStatus)
+	if err != nil {
```
hanzei, line 144: 1/5 `createRepoStatus` should not return an error and be just fire and forget. Because the caller doesn't care most of the time if it succeeded.

server/github.go
hanzei: Are these changes this valuable?

server/github.go
```diff
@@ -300,3 +308,19 @@ func (s *Server) waitForStatus(ctx context.Context, pr *model.PullRequest, statu
 		}
 	}
 }
+
+func (s *Server) GetStatus(pr *model.PullRequest, statusContext string) (status *github.RepoStatus, err error) {
+	ctx, cancel := context.WithTimeout(context.Background(), requestTimeoutGithub)
+	defer cancel()
+	statuses, _, err := s.GithubClient.Repositories.ListStatuses(ctx, pr.RepoOwner, pr.RepoName, pr.Sha, nil)
```
hanzei, line 315: Similar comment as above: Let's just go with a high `PerPage` number instead of proper paging.

server/github.go
```diff
@@ -256,18 +263,19 @@ func (s *Server) areChecksSuccessfulForPr(pr *model.PullRequest, org string) (bo
...
 	mlog.Debug("Retrieved status for pr", mlog.String("status", cStatus.GetState()), mlog.Int("prNumber", pr.Number), mlog.String("prSha", pr.Sha))
-	if cStatus.GetState() == stateSuccess || cStatus.GetState() == "pending" || cStatus.GetState() == "" {
+	if cStatus.GetState() == stateSuccess || cStatus.GetState() == statePending || cStatus.GetState() == "" {
 		return true, nil
 	}
 	return false, nil
 }
 
-func (s *Server) createRepoStatus(ctx context.Context, pr *model.PullRequest, status *github.RepoStatus) {
+func (s *Server) createRepoStatus(ctx context.Context, pr *model.PullRequest, status *github.RepoStatus) error {
```
hanzei, line 272: 1/5 `createRepoStatus` should not return an error and be just fire and forget. Because the caller doesn't care most of the time if it succeeded. Logging to Mattermost can happen here

server/pull_request.go
```diff
@@ -25,6 +25,7 @@ func (s *Server) handlePullRequestEvent(event *PullRequestEvent) {
 	switch event.Action {
 	case "opened":
 		mlog.Info("PR opened", mlog.String("repo", pr.RepoName), mlog.Int("pr", pr.Number))
+		s.createCLAPendingStatus(pr)
 		s.checkCLA(pr)
```
hanzei, line 29: Given that `createCLAPendingStatus` should happen before every `checkCLA`, I think it's worth combining them.

hanzei, line 29: Maybe put this into a separate go routine to not block the rest of the method?

server/pull_request.go
```diff
@@ -433,26 +436,28 @@ func (s *Server) CleanOutdatedPRs() {
...
-		pull, _, errPull := s.GithubClient.PullRequests.Get(context.Background(), pr.RepoOwner, pr.RepoName, pr.Number)
-		if errPull != nil {
-			mlog.Error("Error getting Pull Request", mlog.String("RepoOwner", pr.RepoOwner), mlog.String("RepoName", pr.RepoName), mlog.Int("PRNumber", pr.Number), mlog.Err(errPull))
-			if _, ok := errPull.(*github.RateLimitError); ok {
-				mlog.Error("GitHub rate limit reached")
-				s.CheckLimitRateAndSleep()
-			}
+		pull, _, err := s.GithubClient.PullRequests.Get(ctx, pr.RepoOwner, pr.RepoName, pr.Number)
+		if _, ok := err.(*github.RateLimitError); ok {
+			s.sleepUntilRateLimitAboveTokenReserve()
```
hanzei, line 446: This will sleep for a most an hour. At this point `ctx` got canceled.

With all the CLA changes I'm wondering if these changes are still needed. Maybe it's better to stash them and come back if needed.

server/cla.go
metanerd: To implement it later properly.

server/github.go
metanerd: No need for this method anymore.

server/github.go
metanerd: I would argue that sometimes I do care very much (ee tests and CLA) and I don't want to make too much noise to mattermost when I do not care.

server/pull_request.go
metanerd: Agreed, moved.

server/cla.go
metanerd: Added pagination.

server/cla.go
metanerd: Improved error message.

server/cla.go
metanerd: Yes, nice!

server/pull_request.go
metanerd: This is the cron cleaning up the database at night with outdated PRs. It will just return now, when it runs into the rate limit and not continue cleaning up.

server/cla.go
metanerd: :100:

server/cla.go
metanerd: It's definitely shorter, I will use that.

server/cla.go
metanerd: Out of scope, but yeah this is a long term goal of mine. :+1:

server/config.go
metanerd: I have already contexts in the config. This makes it consistent.

server/enterprise.go
metanerd: yes! I really need to get this PR deployed.

server/enterprise.go
metanerd: I would argue that sometimes I am interested to report it to mattermost. But only on a select few occasions.

server/cla.go
hanzei: Are you still planing to do that?

server/github.go
```diff
@@ -128,13 +131,30 @@ func (s *Server) removeLabel(repoOwner, repoName string, number int, label strin
...
 }
 
-func (s *Server) getComments(repoOwner, repoName string, number int) ([]*github.IssueComment, error) {
-	comments, _, err := s.GithubClient.Issues.ListComments(context.Background(), repoOwner, repoName, number, nil)
-	if err != nil {
-		mlog.Error("pr_error", mlog.Err(err))
-		return nil, err
+func (s *Server) getComments(ctx context.Context, pr *model.PullRequest) (comments []*github.IssueComment, err error) {
+	opts := &github.IssueListCommentsOptions{
+		ListOptions: github.ListOptions{},
```
hanzei, line 136: I've been thinking about reducing the number of request and this seams like another good opportunity: By specifying `PerPage` to `100` this only take a third of the requests.

server/cla.go
metanerd: Done

server/github.go
hanzei: I'm 2/5 that these few requests here don't make a difference. Also, given that the code just keeps the old list of fetching the new fails, it's not worth re-trying.

server/github.go
```diff
@@ -256,18 +280,19 @@ func (s *Server) areChecksSuccessfulForPr(pr *model.PullRequest, org string) (bo
...
 		return true, nil
 	}
 	return false, nil
 }
 
-func (s *Server) createRepoStatus(ctx context.Context, pr *model.PullRequest, status *github.RepoStatus) {
+func (s *Server) createRepoStatus(ctx context.Context, pr *model.PullRequest, status *github.RepoStatus) error {
 	_, _, err := s.GithubClient.Repositories.CreateStatus(ctx, pr.RepoOwner, pr.RepoName, pr.Sha, status)
 	if err != nil {
 		mlog.Error("Unable to create the github status for for PR", mlog.Int("pr", pr.Number), mlog.Err(err))
```
hanzei, line 292: It's not idiomatic to both log and error and return it. The caller should decide what to do with an error.

server/pull_request.go
hanzei: Did you forgot to remove `s.createCLAPendingStatus`?

server/pull_request.go
```diff
@@ -433,26 +435,27 @@ func (s *Server) CleanOutdatedPRs() {
...
-			if _, ok := errPull.(*github.RateLimitError); ok {
-				mlog.Error("GitHub rate limit reached")
-				s.CheckLimitRateAndSleep()
-			}
+		pull, _, err := s.GithubClient.PullRequests.Get(ctx, pr.RepoOwner, pr.RepoName, pr.Number)
+		if _, ok := err.(*github.RateLimitError); ok {
+			return
+		}
+		if err != nil {
+			mlog.Error("Error getting PR", mlog.String("RepoOwner", pr.RepoOwner), mlog.String("RepoName", pr.RepoName), mlog.Int("PRNumber", pr.Number), mlog.Err(err))
```
hanzei, line 448: Should the code `return` here?

server/pull_request.go
metanerd: I did :facepalm:

server/github.go
metanerd: > What happened to the integrations tests?

The function for them was superflous. So I removed the file.

server/github.go
metanerd: :+1:

PR #138, opened 2020-06-05
Succeed CLA status for excluded user.

metanerd:
#### Summary

CLA github status will still be on pending even when excluded bot user. Succeeding the CLA github status is necessary for an excluded user.

### DIFF
--- a/server/cla.go
+++ b/server/cla.go
@@ -52,7 +52,14 @@ func (s *Server) checkCLA(pr *model.PullRequest) {
 	)
 
 	if contains(s.Config.CLAExclusionsList, username) {
-		mlog.Info(fmt.Sprintf("%s is excluded to sign the CLA", username))
+		status := &github.RepoStatus{
+			State:       github.String(stateSuccess),
+			Description: github.String(fmt.Sprintf("%s excluded", username)),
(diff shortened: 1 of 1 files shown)


server/cla.go
```diff
@@ -52,7 +52,14 @@ func (s *Server) checkCLA(pr *model.PullRequest) {
...
 
 	if contains(s.Config.CLAExclusionsList, username) {
-		mlog.Info(fmt.Sprintf("%s is excluded to sign the CLA", username))
+		status := &github.RepoStatus{
+			State:       github.String(stateSuccess),
+			Description: github.String(fmt.Sprintf("%s excluded", username)),
+			TargetURL:   github.String(s.Config.SignedCLAURL),
+			Context:     github.String(s.Config.CLAGithubStatusContext),
+		}
+		mlog.Debug("will succeed CLA status for excluded user", mlog.String("user", username))
```
khos2ow, line 61: Why did you change it from `Info` to `Debug` though?

metanerd: To reduce at some point the noise level in the logs. You want to keep `Info`?

khos2ow: No, just wondering. Looks good with debug.

PR #141, opened 2020-06-24
[MM-26417] Refactor use of context and add timeouts

ethervoid:
#### Summary

- Refactor the use of context
- Create a consistent timeout
- AWS calls uses the `*WithContext` call
- ~Timeouts are configurable~

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-26417

### DIFF
--- a/config/config-mattermod.default.json
+++ b/config/config-mattermod.default.json
@@ -8,7 +8,7 @@
     "CircleCIToken": "",
 
     "DriverName": "mysql",
-    "DataSource": "mattermod:mattermod@tcp(127.0.0.1:3306)/mattermod?parseTime=true",
+    "DataSource": "mattermod:mattermod@tcp(192.168.254.4:3306)/mattermod?parseTime=true",
 
     "TickRateMinutes": 15,
     "SpinmintExpirationHour": 72,
@@ -109,6 +109,11 @@
 
     "MattermostWebhookURL": "",
     "MattermostWebhookFooter": "",
+
+    "RequestTimeoutInSeconds": 60,
+    "CronTaskTimeoutInSeconds": 300,
+    "BuildMobileTaskTimeoutInSeconds": 7200,
+    "BuildSpinmintTaskTimeoutInSeconds": 2700,
 
     "LogSettings": {
         "EnableConsole": true,
--- a/config/config-mattermod.default.json
+++ b/config/config-mattermod.default.json
@@ -8,7 +8,7 @@
     "CircleCIToken": "",
 
     "DriverName": "mysql",
-    "DataSource": "mattermod:mattermod@tcp(192.168.254.4:3306)/mattermod?parseTime=true",
+    "DataSource": "mattermod:mattermod@tcp(127.0.0.1:3306)/mattermod?parseTime=true",
 
     "TickRateMinutes": 15,
     "SpinmintExpirationHour": 72,
--- a/config/config-mattermod.default.json
+++ b/config/config-mattermod.default.json
@@ -111,6 +111,7 @@
     "MattermostWebhookFooter": "",
 
     "RequestTimeoutInSeconds": 60,
+    "EETaskTimeoutInSeconds": 300,
     "CronTaskTimeoutInSeconds": 300,
     "BuildMobileTaskTimeoutInSeconds": 7200,
     "BuildSpinmintTaskTimeoutInSeconds": 2700,
--- a/config/config-mattermod.default.json
+++ b/config/config-mattermod.default.json
@@ -110,12 +110,6 @@
     "MattermostWebhookURL": "",
     "MattermostWebhookFooter": "",
 
-    "RequestTimeoutInSeconds": 60,
-    "EETaskTimeoutInSeconds": 300,
-    "CronTaskTimeoutInSeconds": 300,
-    "BuildMobileTaskTimeoutInSeconds": 7200,
-    "BuildSpinmintTaskTimeoutInSeconds": 2700,
-
     "LogSettings": {
         "EnableConsole": true,
         "ConsoleLevel": "DEBUG",
--- a/server/circleci.go
+++ b/server/circleci.go
@@ -44,7 +44,7 @@ func (s *Server) triggerCircleCiIfNeeded(ctx context.Context, pr *model.PullRequ
 	}
 
 	// List the files that was modified or added in the PullRequest
-	prFiles, _, err := s.GithubClient.PullRequests.ListFiles(context.Background(), pr.RepoOwner, pr.RepoName, pr.Number, nil)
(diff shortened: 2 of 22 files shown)


ethervoid:
@agnivade This is my first iteration on this. I need a bit of help given that is my first time working with contexts and timeouts in Go.

My approach is:

- Create a context for every request received or Cron jobs with a default timeout. Basically we're doing HTTP requests to GitHub so there shouldn't be a problem
- There are two special cases: 
  - `buildMobileApp` which needs it's own [2 hours](https://github.com/mattermost/mattermost-mattermod/blob/master/server/build_mobile_apps.go#L47) context because is a heavy job
  - `waitForBuildAndSetupSpinmint` for the same reason but for [45 minutes](https://github.com/mattermost/mattermost-mattermod/blob/master/server/spinmint.go#L35)
- There was another custom context with [5 minutes timeout](https://github.com/mattermost/mattermost-mattermod/blob/master/server/enterprise.go#L23), enterprise tests, but AFAIK it's only doing requests to Github as well

agnivade:
Looks perfectly fine to me. Just do a little bit of testing to check that once a timeout is over, subsequent requests are indeed cancelled. The codebase doesn't really propagate the error upwards but just logs it and moves on. Something we need to improve.

ethervoid:
@agnivade Would be great a first review until I'm able to test this properly. Doing tests would be very cumbersome because we should mock almost every github call until we reach one of our "context timeout checks" :\

config/config-mattermod.default.json
```diff
@@ -110,6 +110,12 @@
     "MattermostWebhookURL": "",
     "MattermostWebhookFooter": "",
 
+    "RequestTimeoutInSeconds": 60,
+    "EETaskTimeoutInSeconds": 300,
+    "CronTaskTimeoutInSeconds": 300,
+    "BuildMobileTaskTimeoutInSeconds": 7200,
+    "BuildSpinmintTaskTimeoutInSeconds": 2700,
```
agnivade, line 117: I thought as the first cut, we agreed for them to be consts in the code? Unless needed, I would want to avoid more config knobs.

server/pull_request.go
```diff
@@ -113,27 +113,27 @@ func (s *Server) handlePullRequestEvent(event *PullRequestEvent) {
 
 			mlog.Info("test server instance", mlog.String("test server", spinmint.InstanceID))
 			mlog.Info("Will destroy the test server for a merged/closed PR.")
-			s.sendGitHubComment(pr.RepoOwner, pr.RepoName, pr.Number, s.Config.DestroyedSpinmintMessage)
+			s.sendGitHubComment(ctx, pr.RepoOwner, pr.RepoName, pr.Number, s.Config.DestroyedSpinmintMessage)
 			go s.destroySpinmint(pr, spinmint.InstanceID)
```
agnivade, line 117: We should also create a context for the spinmint destruction, in the AWS API calls.

server/circleci.go
```diff
@@ -20,7 +20,7 @@ import (
 	"github.com/metanerd/go-circleci"
 )
 
-func (s *Server) triggerCircleCiIfNeeded(pr *model.PullRequest) {
+func (s *Server) triggerCircleCiIfNeeded(ctx context.Context, pr *model.PullRequest) {
 	client := &circleci.Client{Token: s.Config.CircleCIToken}
```
agnivade, line 24: We also need to pass timeouts to the circleCI calls. The metanerd/go-circleci package does not expose context in the API calls. I am 0/5 on whether we want to bring the package inside this repo or modify it upstream. But let's file a ticket to pass the context.

config/config-mattermod.default.json
ethervoid: Those are optional, we're creating the default when the config is loaded. But this provides flexibility if we need to tune the timeouts in production with just a restart instead of making a deployment. This way we should be able to change timeouts if the default ones are not long enough.

I'll remove the timeouts in the configuration but I'll leave it configurable as a "private" feature right now. What do you think?

server/pull_request.go
ethervoid: You mean check for `ctx.Done()` in the `destroySpinmint` method right? AWS API calls don't accept context

server/circleci.go
ethervoid: That project is also a fork from Carlos's fork :laughing: . I'll ask here to open issues and fill one asking for exposing the context in the calls

agnivade: Thanks :)

config/config-mattermod.default.json
agnivade: Ok, but then let's change the fields to be time.Duration instead of int. It feels odd to have getters which convert int to time.Duration every time we call it.

ethervoid: Hum but in that case, if we introduce those fields in the configuration file are going to be integers and the JSON parse is going to fail.

server/circleci.go
ethervoid: Create https://mattermost.atlassian.net/browse/MM-26589

server/pull_request.go
agnivade: I meant the AWS API calls. `destroySpinmint` happens in it's own goroutine so that need not take the same context.

From what I can see every AWS API has a `WithContext` counterpart. For example: https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#EC2.RunInstancesWithContext.

used in server/spinmint.go:L150.

Is there any API in particular that doesn't take context?

ethervoid: Ok, I misunderstood the point. :+1:

config/config-mattermod.default.json
agnivade: That's a good point, but isn't too hard to achieve: https://play.golang.org/p/Zm6hpNR-ZJ2.

server/pull_request.go
ethervoid: Added

config/config-mattermod.default.json
ethervoid: Finally, I've removed the configuration complexity

agnivade:
Btw, the `gojsonschema` should have been removed from go.mod in https://github.com/mattermost/mattermost-mattermod/commit/7028b1b6cf4ed7f0c28cbdb224e5f46a9c945cd5. /cc @isacikgoz. We need a "go mod tidy" CI check in the repo. I'll create a ticket.

ethervoid:
I’ve applied go mod tidy [here](https://github.com/mattermost/mattermost-mattermod/pull/141/commits/b3b1e91e6a17c44a3d3c4af1bf92314329ec5102)

agnivade:
Yeah that's perfect. I was just saying that it should have been applied in that PR I linked above, and not caught so late.

PR #143, opened 2020-06-25
Reimplement translation prs

metanerd:
#### Summary
Before the migration from pootle to weblate, we got translation PRs for review into the i18n channel. This PR implements the same logic for weblate.


server/webhook.go
```diff
@@ -18,14 +20,20 @@ type Payload struct {
 	Text     string `json:"text"`
 }
 
-func (s *Server) sendToWebhook(payload *Payload) error {
+func (s *Server) sendToWebhook(webhookURL string, payload *Payload) error {
+	if webhookURL == "" {
+		err := errors.New("no Mattermost webhook URL set: unable to send message")
```
khos2ow, line 25: The same, may be drop `Mattermost`?

server/translations.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+func (s *Server) handleTranslationPr(pr *model.PullRequest) {
+	if pr.Username != s.Config.Username {
+		return
+	}
+
+	dataMsg := fmt.Sprintf("####[%v translations PR %v](%v)\n", pr.RepoName, time.Now().UTC().Format(time.RFC3339), pr.URL)
+	msg := dataMsg + s.Config.TranslationsMattermostMessage
+	mlog.Debug("Sending Mattermost message", mlog.String("message", msg))
```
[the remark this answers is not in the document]
metanerd: Why does it make it clearer?

server/enterprise.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 func (s *Server) triggerEETestsForOrgMembers(pr *model.PullRequest) {
-	if s.IsOrgMember(pr.Username) {
+	if s.IsOrgMember(pr.Username) || s.IsKnownUser(pr.Username) {
```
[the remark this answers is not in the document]
metanerd: I renamed it to IsTrustedUser(). I am for adding this, because the translation bot only modifies translation files and otherwise I have to add the EETests label and then check in again in an hour. But if you have objections, I am fine with not whitelisting the CLAExcluded bots.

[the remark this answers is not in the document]
khos2ow: No, I'm fine and there's no harm running EE tests anyway.

server/translations.go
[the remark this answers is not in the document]
khos2ow: Because the payload is being sent to translation service and not Mattermost. Isn't it?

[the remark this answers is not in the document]
metanerd: It is sent to mattermost. It contains the links to the PRs for translations: https://community.mattermost.com/core/pl/sqzth7ryibgrbqj196trrs4hec

server/translations.go
```diff
@@ -0,0 +1,27 @@
...
+
+	webhookRequest := &Payload{Username: "Weblate", Text: msg}
+	r, err := s.sendToWebhook(s.Config.TranslationsMattermostWebhookURL, webhookRequest)
+	if err != nil {
+		mlog.Error("Unable to post to Mattermost webhook", mlog.Err(err))
+		return
+	}
+	defer r.Body.Close()
```
agnivade, line 26: Any reason to use `defer` as the last statement of a function? Or do we expect more lines to be added later?

Also, please nil-check `Body` before calling `Close`.

Lastly, let's read the body to completion to reuse the same connection.

server/config.go
```diff
@@ -99,6 +99,10 @@ type Config struct {
 	EnterpriseGithubStatusEETests string
 	EnterpriseWorkflowName        string
 
+	TranslationsMattermostWebhookURL string
+	TranslationsMattermostMessage    string
+	TranslationsBot                  string
```
agnivade, line 104: This field is unused. Any reason for that?

server/github.go
```diff
@@ -212,6 +212,15 @@ func (s *Server) IsOrgMember(user string) bool {
 	return false
 }
 
+func (s *Server) IsTrustedUser(user string) bool {
```
agnivade, line 215: godocs please :)

server/utils.go
```diff
@@ -21,9 +16,12 @@ func (s *Server) logToMattermost(msg string, args ...interface{}) {
 
 	webhookRequest := &Payload{Username: "Mattermod", Text: webhookMessage}
 
-	if err := s.sendToWebhook(webhookRequest); err != nil {
+	r, err := s.sendToWebhook(s.Config.MattermostWebhookURL, webhookRequest)
+	if err != nil {
 		mlog.Error("Unable to post to Mattermost webhook", mlog.Err(err))
+		return
 	}
+	defer r.Body.Close()
```
agnivade, line 24: Same comment as earlier.

server/webhook.go
```diff
@@ -18,31 +17,55 @@ type Payload struct {
...
 }
 
-func (s *Server) sendToWebhook(payload *Payload) error {
+func (s *Server) sendToWebhook(webhookURL string, payload *Payload) (*http.Response, error) {
+	err := validateSendToWebhookRequest(webhookURL, payload)
+	if err != nil {
+		badRequestR := &http.Response{
+			StatusCode: http.StatusBadRequest,
+			Body:       ioutil.NopCloser(bytes.NewBufferString(err.Error())),
+		}
```
agnivade, line 26: I am wondering the reasoning behind this? The body only contains the error string, and we are already returning the error. What extra are we including by injecting an artificially constructed response?

Same for others below.

server/webhook_integration_test.go
```diff
@@ -0,0 +1,81 @@
+package server
```
agnivade, line 1: files are typically named by appending "_test" to the file it is testing. So in this case, I would suggest "webhook_test.go".

server/webhook_integration_test.go
```diff
@@ -0,0 +1,81 @@
...
+	defer r.Body.Close()
+}
+
+func TestSendToWebhookUsernameNotSet(t *testing.T) {
+	invalidPayload := &Payload{Text: "test"}
+	mattermost := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+	}))
+	defer mattermost.Close()
+
+	wantError := "username not set in webhook payload"
```
agnivade, line 35: String matching in errors is brittle and not recommended. We can name the error as a variable if we want to verify that. Or had a separate type like 'webhookValidationError' with field set to Username, Text etc. Up to you.

Same applies to the test below too.

server/webhook_integration_test.go
```diff
@@ -0,0 +1,81 @@
...
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
+func TestSendToWebhook(t *testing.T) {
+	validPayload := &Payload{Username: "mattermod", Text: "test"}
+
+	mattermost := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		fmt.Fprintln(w, "ok")
```
agnivade, line 14: Please also validate the payload sent.

server/webhook_integration_test.go
```diff
@@ -0,0 +1,81 @@
...
+	}))
+	defer mattermost.Close()
+
+	r, err := s.sendToWebhook(mattermost.URL, validPayload)
+	if err != nil {
+		t.Fatal(err.Error())
```
agnivade, line 20: We use testify as our testing framework. Let's use that here. 

Same applies to everywhere else.

server/webhook_integration_test.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+	got := r.StatusCode
+	if got != http.StatusOK {
+		t.Errorf("got: %v, want: %v", got, http.StatusOK)
+	}
+	defer r.Body.Close()
```
agnivade, line 26: Same comment as earlier.

server/enterprise.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 func (s *Server) triggerEETestsForOrgMembers(pr *model.PullRequest) {
-	if s.IsOrgMember(pr.Username) {
+	if s.IsOrgMember(pr.Username) || s.IsTrustedUser(pr.Username) {
```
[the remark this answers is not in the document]
metanerd: TrustedUsers refers to for example dependabot or weblate bot. I tried to use a generic name. I can also rename them to bots again, but it boils down to them being github users that we trust IMO. Well, it is debatable.

server/config.go
metanerd: Nice catch! Totally messed up there.

server/enterprise.go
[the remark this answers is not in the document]
agnivade: Ah I see. Are there actual users in the list apart from bots?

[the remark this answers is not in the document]
metanerd: No, no users. Only bots.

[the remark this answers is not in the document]
agnivade: Thanks for the clarification. Perhaps we can look into a better name later.

server/cla.go
```diff
@@ -103,21 +103,25 @@ func (s *Server) checkCLA(ctx context.Context, pr *model.PullRequest) {
...
+	if err != nil {
+		return nil, err
+	}
+	r, err := http.DefaultClient.Do(req)
 	if err != nil {
 		mlog.Error("Unable to get CLA list", mlog.Err(err))
-		s.logToMattermost("unable to get CLA google csv file Error: ```" + err.Error() + "```")
+		s.logToMattermost(ctx, "unable to get CLA google csv file Error: ```"+err.Error()+"```")
 		return nil, err
 	}
```
agnivade, line 116: `closeBody` should be called after this, not after `ioutil.ReadAll`.

server/utils.go
agnivade: I am not seeing this resolved. Please see my comment https://github.com/mattermost/mattermost-mattermod/pull/143/#discussion_r446847856. We can just use `closeBody` here.

server/webhook.go
agnivade: This does not appear to be resolved?

server/webhook.go
```diff
@@ -5,44 +5,75 @@ package server
...
 )
 
+type WebhookValidationError struct {
+	err string
+}
+
+func (e *WebhookValidationError) Error() string {
+	return fmt.Sprintf("%v", e.err)
+}
+
```
agnivade, line 22: This implements only partially what I suggested. This just wraps the error with a type, and does not do anything else. As it is, it's not very useful. I was suggesting that a field added to the struct which will indicate what is not set- url/username/text.

Please also add godocs to both the struct and the method.

server/cla.go
```diff
@@ -103,21 +103,25 @@ func (s *Server) checkCLA(ctx context.Context, pr *model.PullRequest) {
...
-func (s *Server) getCSV() ([]byte, error) {
-	resp, err := http.Get(s.Config.SignedCLAURL)
+func (s *Server) getCSV(ctx context.Context) ([]byte, error) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, s.Config.SignedCLAURL, http.NoBody)
+	if err != nil {
+		return nil, err
+	}
+	r, err := http.DefaultClient.Do(req)
 	if err != nil {
 		mlog.Error("Unable to get CLA list", mlog.Err(err))
```
agnivade, line 113: Unrelated to this PR, but this is an anti-pattern. Logging and then returning the same error. We need to do a sweep to fix all of these.

server/webhook_test.go
```diff
@@ -0,0 +1,88 @@
...
+	"context"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
```
agnivade, line 13: Nit: remove gap

server/webhook_test.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestSendToWebhookIntegration(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping integration test")
+	}
```
agnivade, line 20: Any reason for this?

Same for below.

server/webhook_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+	s := &Server{
+		StartTime: time.Now(),
```
agnivade, line 23: Why do we need this in the test?

Same for below.

server/webhook_test.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	validPayload := &Payload{Username: "mattermod", Text: "test"}
+	expectedMattermostPayload := "ok"
+	mattermost := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		_, err := w.Write([]byte(expectedMattermostPayload))
+		if err != nil {
+			log.Fatal(err)
+		}
```
agnivade, line 32: I wanted to validate the incoming payload. Not the outgoing response. :)

Validating that we are returning `ok` is fine but is not very important. It's a response from an external server. But our objective is to validate our code. i.e. we want to verify that we are sending the correct data to the server. Let's do that.

server/webhook_test.go
```diff
@@ -0,0 +1,88 @@
...
+	if testing.Short() {
+		t.Skip("skipping integration test")
+	}
+
+	s := &Server{
+		StartTime: time.Now(),
+	}
+
+	validPayload := &Payload{Username: "mattermod", Text: "test"}
+	expectedMattermostPayload := "ok"
```
agnivade, line 27: I don't think there's a need to do this. But if you want it, I'll suggest to just use a byte slice. That'll prevent the double-allocation to strings.

server/webhook_test.go
```diff
@@ -0,0 +1,88 @@
...
+		StartTime: time.Now(),
+	}
+
+	invalidPayload := &Payload{Text: "test"}
+	mattermost := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+	}))
+	defer mattermost.Close()
+
+	r, err := s.sendToWebhook(context.Background(), mattermost.URL, invalidPayload)
+	require.Error(t, err.(*WebhookValidationError), err)
```
agnivade, line 61: We should use the `errors.As` function to map to an error type and then check for true/false. The current method will fail if the error is wrapped.

Same for below.

server/webhook_test.go
metanerd: We are calling `s.sendToWebhook(context.Background(), mattermost.URL, validPayload)`. Is there another way?

server/webhook_test.go
metanerd: I was actually confused. I didn't want to validate that payload.

server/webhook.go
metanerd: Can you please send me an example for what you mean by this?

server/webhook_test.go
agnivade: Sorry, I meant the `StartTime` field. Is that needed?

server/webhook_test.go
metanerd: I want to differentiate between running unit tests and implementation tests. I looked up on how to accomplish this. This was, what came up as a go to solution.

server/webhook.go
agnivade: Sure.

```go
// WebhookValidationError contains an error in the webhook payload.
type WebhookValidationError struct {
 field string
}

// Error implements the error interface.
func (e *WebhookValidationError) Error() string {
	return "invalid" + e.field
}

func newWebhookValidationError(field string) *WebhookValidationError {
  return &WebhookValidationError{
   field: field,
  }
}

// from code
if webhookURL == "" {
	return newWebhookValidationError("webhook URL")
}

// in test
var whError *WebhookValidationError
require.True(errors.As(err, &whError))
assert.Equal(whError.field, "webhook URL")
```

Let me know if that helps. (Code is untested, I just wrote here directly)

metanerd: This helps immensely! Thank you so much!

server/webhook_test.go
agnivade: But this is an unit test from what I understand. There is no external dependency with any other service. It's purely inside the code.

metanerd: OK, removed.

server/webhook.go
```diff
@@ -35,14 +53,22 @@ func (s *Server) sendToWebhook(payload *Payload) error {
...
+	if webhookURL == "" {
+		return newWebhookValidationError("webhook URL")
+	}
+
+	if payload.Username == "" {
+		return newWebhookValidationError("username")
 	}
 
+	if payload.Text == "" {
+		return &WebhookValidationError{"text"}
```
agnivade, line 71: move to new style?

server/webhook.go
```diff
@@ -5,27 +5,45 @@ package server
...
 )
 
+// WebhookValidationError contains an error in the webhook payload.
+type WebhookValidationError struct {
+	field string
+}
+
+// Error implements the error interface.
+func (e *WebhookValidationError) Error() string {
+	return "invalid" + e.field
```
agnivade, line 20: space after "invalid"?

server/cla.go
```diff
@@ -103,19 +103,21 @@ func (s *Server) checkCLA(ctx context.Context, pr *model.PullRequest) {
...
-		s.logToMattermost("unable to get CLA google csv file Error: ```" + err.Error() + "```")
 		return nil, err
 	}
-	defer resp.Body.Close()
+	r, err := http.DefaultClient.Do(req)
+	if err != nil {
+		s.logToMattermost(ctx, "unable to get CLA google csv file Error: ```"+err.Error()+"```")
+		return nil, err
+	}
+	closeBody(r)
```
agnivade, line 116:
```suggestion
	defer closeBody(r)
```

metanerd:
Thanks to you, sir!

PR #144, opened 2020-06-29
Blacklist files per repository

hanzei:
#### Summary
Instead of a global list the file blacklist should be per repo. 

I didn't add a unit test as there is currently no way to mock the circleci client.

#### Ticket Link
https://mattermost.atlassian.net/browse/DOPS-136


server/circleci.go
```diff
@@ -43,20 +44,26 @@ func (s *Server) triggerCircleCiIfNeeded(pr *model.PullRequest) {
...
-				mlog.Error("File is on the blacklist and will not retrigger circleci to give the contexts", mlog.String("repo", pr.RepoName), mlog.Int("pr", pr.Number), mlog.String("Fullname", pr.FullName))
-				msg := fmt.Sprintf("The file `%s` is in the blacklist and should not be modified from external contributors, please if you are part of the Mattermost Org submit this PR in the upstream.\n /cc @mattermost/core-security @mattermost/core-build-engineers", prFile.GetFilename())
-				s.sendGitHubComment(pr.RepoOwner, pr.RepoName, pr.Number, msg)
-				return
+		for _, blackListItem := range blackList {
+			for _, prFile := range prFiles {
+				if prFile.GetFilename() == blackListItem {
+					mlog.Error("File is on the blacklist and will not retrigger circleci to give the contexts", mlog.String("repo", pr.RepoName), mlog.Int("pr", pr.Number), mlog.String("Fullname", pr.FullName))
+					msg := fmt.Sprintf(
+						"The file `%s` is in the blacklist for external contributors. Hence, these changes are not tested by the CI pipeline active until the PR is merged. Please be careful when reviewing it.\n /cc @mattermost/core-security @mattermost/core-build-engineers",
```
jupenur, line 62: > changes are not tested by the CI pipeline active until the PR is merged

This is inaccurate. An org member can trigger the checks for the contributor afaik, and then they should run just fine.

hanzei:
Why should `.circleci/config.yml` get blacklisted for repositories without any context?

jupenur:
@hanzei because that's where the contexts are assigned. There are also a couple of other entries that need to remain global, but I won't name them here in public for reasons.

metanerd:
Requesting a review by @agnivade because of mattermod revamp.

server/circleci.go
```diff
@@ -43,20 +44,26 @@ func (s *Server) triggerCircleCiIfNeeded(pr *model.PullRequest) {
...
-	// List the files that was modified or added in the PullRequest
-	prFiles, _, err := s.GithubClient.PullRequests.ListFiles(context.Background(), pr.RepoOwner, pr.RepoName, pr.Number, nil)
-	if err != nil {
-		mlog.Error("Error listing the files from a PR", mlog.String("repo", pr.RepoName), mlog.Int("pr", pr.Number), mlog.String("Fullname", pr.FullName), mlog.Err(err))
-		return
-	}
+	blackList, ok := s.Config.FileBlacklist[pr.RepoName]
+	if ok {
+		// List the files that was modified or added in the PullRequest
+		var prFiles []*github.CommitFile
```
agnivade, line 50: Any reason this needs to be declared as a separate variable?

server/circleci.go
```diff
@@ -43,20 +44,26 @@ func (s *Server) triggerCircleCiIfNeeded(pr *model.PullRequest) {
...
-		for _, blackListPath := range s.Config.BlacklistPaths {
-			if prFile.GetFilename() == blackListPath {
-				mlog.Error("File is on the blacklist and will not retrigger circleci to give the contexts", mlog.String("repo", pr.RepoName), mlog.Int("pr", pr.Number), mlog.String("Fullname", pr.FullName))
-				msg := fmt.Sprintf("The file `%s` is in the blacklist and should not be modified from external contributors, please if you are part of the Mattermost Org submit this PR in the upstream.\n /cc @mattermost/core-security @mattermost/core-build-engineers", prFile.GetFilename())
-				s.sendGitHubComment(pr.RepoOwner, pr.RepoName, pr.Number, msg)
-				return
+		for _, blackListItem := range blackList {
+			for _, prFile := range prFiles {
+				if prFile.GetFilename() == blackListItem {
+					mlog.Error("File is on the blacklist and will not retrigger circleci to give the contexts", mlog.String("repo", pr.RepoName), mlog.Int("pr", pr.Number), mlog.String("Fullname", pr.FullName))
```
agnivade, line 60: Perhaps (re)trigger is better?

server/circleci.go
metanerd: There needs to be a separate PR for platform-private/mattermod.

agnivade: Did you mean to make this comment on my parent message or the one you replied to?

metanerd: > If this gets auto-deployed when pushed to master, it will break because it expects the config to be in a different way. How do we plan to handle this? @metanerd

To this one.

hanzei:
@agnivade @metanerd @jupenur I've updated the PR to be backward compatible. PTAL

server/circleci.go
```diff
@@ -49,11 +49,17 @@ func (s *Server) triggerCircleCiIfNeeded(ctx context.Context, pr *model.PullRequ
...
-		for _, blockListPath := range s.Config.BlacklistPaths {
+		for _, blockListPath := range blockList {
 			if prFile.GetFilename() == blockListPath {
-				mlog.Error("File is on the blocklist and will not retrigger circleci to give the contexts", mlog.String("repo", pr.RepoName), mlog.Int("pr", pr.Number), mlog.String("Fullname", pr.FullName))
-				msg := fmt.Sprintf("The file `%s` is in the blocklist and should not be modified from external contributors, please if you are part of the Mattermost Org submit this PR in the upstream.\n /cc @mattermost/core-security @mattermost/core-build-engineers", prFile.GetFilename())
+				mlog.Error("File is on the blocklist and will not re-trigger circleci to give the contexts", mlog.String("repo", pr.RepoName), mlog.Int("pr", pr.Number), mlog.String("Fullname", pr.FullName))
+				msg := fmt.Sprintf("The file `%s` is in the blacklist for external contributors. Hence, these changes are not tested by the CI pipeline active until the build is re-triggered by a core committer or the PR is merged. Please be careful when reviewing it.\n /cc @mattermost/core-security @mattermost/core-build-engineers", prFile.GetFilename())
```
agnivade, line 62: "blocklist" to be consistent with the codebase.

server/circleci.go
```diff
@@ -49,11 +49,17 @@ func (s *Server) triggerCircleCiIfNeeded(ctx context.Context, pr *model.PullRequ
 		return
 	}
 
+	blockList := s.Config.BlacklistPaths
```
agnivade, line 52: To clarify, does this repo level list act as an override to the global list? Or is the global list supposed to go away? If it's the latter, then it might create considerable bloat in the config because we would need to add a block list for every repo that is tracked.

I am assuming this acts only as an override.

hanzei: As requested by @jupenur the global list should be kept as there are certain files that are blocked in every repo. The repo specific list is an extension of the global list and the union of both is used.

metanerd:
Can we adapt the config for this PR https://github.com/mattermost/platform-private/pull/265  and get it through a security review?

server/circleci.go
agnivade: Right, thanks.

config/config-mattermod.default.json
```diff
@@ -83,6 +83,7 @@
     "SignedCLAURL": "",
     "NeedsToSignCLAMessage": "",
     "BlacklistPaths": [],
```
metanerd, line 85: Maybe to make it more clear we could name it something like BlacklistPathsBase?

server/circleci.go
```diff
@@ -49,11 +49,17 @@ func (s *Server) triggerCircleCiIfNeeded(ctx context.Context, pr *model.PullRequ
...
+	repoBlockList, ok := s.Config.RepoBlockListPaths[pr.RepoName]
+	if ok {
+		blockList = append(blockList, repoBlockList...)
```
metanerd, line 55: :thinking: Can I also get privately messaged the list of files please?

config/config-mattermod.default.json
hanzei: I was asked by @jupenur to keep this PR backwards compatible. I'm 0/5 on this topic.

jupenur: I don't care about the naming in particular, but @agnivade had some concerns regarding deployment.

agnivade: I initially thought that the new key would replace the old one, which would have been a compatibility break. But Elisabeth is suggesting to rename the config param, which is a different thing, but yet another compatiblity break. I am 0/5 too.

hanzei: I think it's worth the naming cleanup here. `BlockListPathsGlobal` and `BlockListPathsPerRepo` is want I went with.

server/circleci.go
hanzei: You mean the new list?  See https://github.com/mattermost/platform-private/pull/265 for that.

hanzei:
I've updated my branch with the changes from https://github.com/mattermost/mattermost-mattermod/pull/215 and added a test.

@metanerd is this PR ready to get merged and deployed?

metanerd:
Yes, gonna deploy now!

PR #145, opened 2020-06-29
[MM-26418] Extract AWS Session creation

ethervoid:
#### Summary

This PR changes how the AWS session is created. Now is created with the server instance and is located in the Server struct to avoid creating a new session every time

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-26418

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -36,7 +36,7 @@ type Server struct {
 	Builds               buildsInterface
 	commentLock          sync.Mutex
 	StartTime            time.Time
-	AWSSession           *session.Session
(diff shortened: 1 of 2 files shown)


server/server.go
```diff
@@ -35,6 +36,7 @@ type Server struct {
 	Builds               buildsInterface
 	commentLock          sync.Mutex
 	StartTime            time.Time
+	AWSSession           *session.Session
```
agnivade, line 39: Any reason to keep this exported?

ethervoid: No at all. Changed

PR #146, opened 2020-06-29
[MM26383] server/server_test: remove globals and slim down tests

isacikgoz:
#### Summary
This PR removes global state form `server_test` and also slims down the test to actually test the `ping()` endpoint. There were too much initialization and complexity for only testing a single endpoint. If server is need to be tested, it should be handled in a different test.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26383

### DIFF
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -4,82 +4,43 @@
 package server
 
 import (
-	"flag"
 	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
-	"os"
 	"path/filepath"
-	"strings"
 	"testing"
+	"time"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
-	"github.com/mattermost/mattermost-server/v5/mlog"
 	"github.com/xeipuuv/gojsonschema"
 )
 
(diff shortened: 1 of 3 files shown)


server/server_test.go
```diff
@@ -4,82 +4,45 @@
...
 	bytesLoader := gojsonschema.NewBytesLoader(bytes)
 
 	dir, err := filepath.Abs(filepath.Dir(""))
-	if err != nil {
-		mlog.Error("unable to find project path")
-	}
-	jsonLoader := gojsonschema.NewReferenceLoader("file://" + dir + "/schema/ping.schema.json")
+	assert.NoError(t, err)
 
+	jsonLoader := gojsonschema.NewReferenceLoader("file://" + dir + "/schema/ping.schema.json")
```
agnivade, line 39: I thought the ticket was also about removing this pingschema.json. Any reason this is here? We should do a json unmarshal and verify the contents.

isacikgoz: I thought the test is more about *at least* ensuring the response is in a certain schema. 🤷 Updated the PR anyway..

PR #148, opened 2020-06-30
[MM-26380] Replace manners with standard library shutdown

isacikgoz:
#### Summary
This PR remove `manners` for `http.Server` shutdown. Also moved `rand.Seed()` into more convenient place. Did some cleanup in the server struct.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26380

### DIFF
--- a/mattermod.go
+++ b/mattermod.go
@@ -32,15 +32,15 @@ func main() {
 
 	mlog.Info("Starting Mattermod Server")
 	go func() {
-		if err := s.Start(); err != nil {
-			mlog.Error("Server exited with error", mlog.Err(err))
+		if err2 := s.Start(); err2 != nil {
+			mlog.Error("Server exited with error", mlog.Err(err2))
 			os.Exit(1)
 		}
 	}()
 	defer func() {
 		mlog.Info("Stopping Mattermod Server")
-		if err := s.Stop(); err != nil {
-			mlog.Error("Error while shutting down server", mlog.Err(err))
+		if err2 := s.Stop(); err2 != nil {
+			mlog.Error("Error while shutting down server", mlog.Err(err2))
 			os.Exit(1)
 		}
 	}()
--- a/mattermod.go
+++ b/mattermod.go
@@ -31,12 +31,8 @@ func main() {
 	s := server.New(config)
 
 	mlog.Info("Starting Mattermod Server")
-	go func() {
-		if err2 := s.Start(); err2 != nil {
-			mlog.Error("Server exited with error", mlog.Err(err2))
-			os.Exit(1)
-		}
-	}()
+	s.Start()
+
 	defer func() {
 		mlog.Info("Stopping Mattermod Server")
 		if err2 := s.Stop(); err2 != nil {
--- a/mattermod.go
+++ b/mattermod.go
@@ -4,11 +4,13 @@
 package main
 
 import (
+	"context"
 	"flag"
 	"fmt"
 	"os"
 	"os/signal"
 	"syscall"
+	"time"
 
 	"github.com/mattermost/mattermost-mattermod/server"
 	"github.com/mattermost/mattermost-server/v5/mlog"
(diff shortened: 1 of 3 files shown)


mattermod.go
```diff
@@ -23,21 +25,16 @@ func main() {
 
 	config, err := server.GetConfig(configFile)
 	if err != nil {
-		err = errors.Wrap(err, "unable to load server config")
-		mlog.Error("unable to load server config", mlog.Err(err))
-		panic(err)
```
agnivade, line 28: :heart:

mattermod.go
```diff
@@ -23,21 +25,16 @@ func main() {
...
+		mlog.Error("unable to load server config", mlog.Err(errors.Wrap(err, "unable to load server config")))
```
agnivade, line 28: Wondering if it makes sense to just do: `mlog.Error("unable to load server config", mlog.Err(err))`? The string is repeated, and we are not really propagating the error upwards.

server/server.go
```diff
@@ -75,34 +73,37 @@ func New(config *Config) (server *Server, err error) {
...
+	errc := make(chan error)
 	go func() {
 		mlog.Info("Listening on", mlog.String("address", s.Config.ListenAddress))
-		err := manners.ListenAndServe(s.Config.ListenAddress, handler)
-		if err != nil {
-			s.logToMattermost(err.Error())
-			mlog.Critical("server_error", mlog.Err(err))
-			panic(err.Error())
-		}
+		errc <- s.server.ListenAndServe()
```
agnivade, line 98: Any reason for such a complicated implementation? `ListenAndServe` will _always_ return an error when stopped via `Shutdown`.

This means there will be deadlock now because `errc` cannot take the error since it's inside the signal handler case.

I thought we would just use the standard approach here: https://golang.org/pkg/net/http/#Server.Shutdown and just do an mlog.Error if it's not a `ErrServerClosed`.

mattermod.go
```diff
@@ -74,5 +71,18 @@ func main() {
...
+
+	select {
+	case <-sig:
+		mlog.Info("Stopping Mattermod Server")
+		ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+		defer cancel()
+		if err := s.Stop(ctx); err != nil {
+			mlog.Error("Error while shutting down server", mlog.Err(err))
+			os.Exit(1)
+		}
```
agnivade, line 83: IMO, we can move it inside `s.Stop`, but 0/5. Up to you.

isacikgoz: I can move the context snippet inside `s.Stop` 👍

server/server.go
isacikgoz: Hmm, actually we are implicitly ignoring `ErrServerClosed` because the interrupt signal will win in the select block in the `main` function. Otherwise we will receive errors from `ListenAndServe`. But the standard way is always more preferable :)

mattermod.go
```diff
@@ -23,21 +22,28 @@ func main() {
...
 
-	s, err := server.New(config)
-	if err != nil {
-		panic("failed creating server")
-	}
-
-	s.Start()
-	defer s.Stop()
+	mlog.Info("Starting Mattermod Server")
+	go func() {
```
agnivade, line 34: 0/5, but it felt cleaner to spawn the goroutine internally from inside `s.Start`. I know there are guidelines that say making calls async should be left to caller. But in general, starting a server is always async from what I have seen.

isacikgoz: Actually I just didn't wanted to `os.Exit()` in the `server.go` :)

agnivade: I don't think you need to. The process is already on the way to its exit. We just need to log it. That's all. Unless I have missed something.

isacikgoz: Let's say if the port is used by some other process, the app will hang in there and will need to be interrupted.

agnivade: That's a good point. Personally, I am fine with os.Exit in server.go. But 0/5, up to you. This is already good to go from my side.

isacikgoz: I see, let's go with your suggestion. I will re-consider this in the `main()` overhaul 👍

PR #150, opened 2020-07-01
[MM-26612] Include go mod tidy check

ethervoid:
#### Summary

This PR includes a new CircleCI check to detect unused dependencies in the go modules and ask for a `go mod tidy` execution

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-26612


ethervoid:
Yes! If you see the commits there are two failing and sure that I included one called “go mod tidy” passed that includes the changes that fixes the problem

agnivade:
Perfect. Thanks.

.circleci/config.yml
```diff
@@ -25,6 +25,17 @@ jobs:
...
 
+  check-dependencies:
+    executor:
+      name: default
+    steps:
+      - checkout
+      - run:
+          name: Check unused Go dependencies
+          command: |
+            go mod tidy
```
hanzei, line 36: Using `-v` will help catching odd situations like a broken connection to the internet.
```suggestion
            go mod tidy -v
```

.circleci/config.yml
```diff
@@ -25,6 +25,17 @@ jobs:
...
+            if [[ -n $(git status --porcelain) ]]; then echo "There are unused dependencies that should be removed. Please execute `go mod tidy` to fix it."; exit 1; fi
```
hanzei, line 37: Nit pick: That check might cause false positives if any other file was modified and also doesn't show what the diff is. What do you think about using `git --no-pager diff --exit-code go.mod go.sum`?

.circleci/config.yml
ethervoid: I was thinking about including it but didn’t now that so perfect :)

.circleci/config.yml
ethervoid: No problem including that to see the diff. The other problem is not going to happen because every step in circleci is running in an isolated environment without possible changes as you can see [here](https://app.circleci.com/pipelines/github/mattermost/mattermost-mattermod/681/workflows/10659967-533e-4822-b948-ae6961b1eef0/jobs/1684/steps)

agnivade:
Thanks for chiming in @hanzei ! Much appreciated.

PR #152, opened 2020-07-02
DOPS-140: Lead translators towards translation server

metanerd:
#### Summary
In the discussion starting [here](https://community.mattermost.com/core/pl/uej5oojr978zjxa7z9wqqbnuah) it turns out we need a way to guide translators towards the translation server, so that translation files are not touched directly via PRs. 
The approach is that as soon as someone creates a PR modifying the language files other than the base language file, there is a message to guide the contributor towards the translation server and the PR gets marked as "Do Not Merge" and the first lifecycle of stale. 
https://github.com/mattermost/platform-private/pull/266

#### Ticket Link

https://mattermost.atlassian.net/jira/software/projects/DOPS/boards/66/backlog?selectedIssue=DOPS-140


server/config.go
```diff
@@ -111,6 +116,7 @@ type Config struct {
 	TranslationsMattermostWebhookURL string
 	TranslationsMattermostMessage    string
 	TranslationsBot                  string
+	TranslationsRepos                []*TranslationsRepo
```
agnivade, line 119: The field name and the JSON key in config are different. We need to fix one of these.

server/config.go
```diff
@@ -55,6 +55,11 @@ type BuildMobileAppJob struct {
 	ExpectedArtifacts int
 }
 
+type TranslationsRepo struct {
```
agnivade, line 58: godocs please :)

server/translations.go
```diff
@@ -3,8 +3,12 @@ package server
...
 	"context"
 	"fmt"
+	"path"
+	"path/filepath"
 	"time"
 
+	"github.com/google/go-github/v32/github"
+
 	"github.com/mattermost/mattermost-mattermod/model"
 	"github.com/mattermost/mattermost-server/v5/mlog"
```
agnivade, line 13: Nit: the import groups are mixed up. They should be arranged in the order of standard library, mattermod, and then 3rd party.

server/translations.go
```diff
@@ -25,3 +29,83 @@ func (s *Server) handleTranslationPR(ctx context.Context, pr *model.PullRequest)
...
+	}
+
+	mlog.Info("Modification of language file", mlog.String("repo", pr.RepoName), mlog.Int("pr", pr.Number), mlog.String("username", pr.Username))
+	msg := fmt.Sprintf("Thank you for your contribution, @%s! For translating, please use [https://translate.mattermost.com](https://translate.mattermost.com) and join our community on [https://community.mattermost.com/core/channels/localization](https://community.mattermost.com/core/channels/localization) ", pr.Username)
+	s.sendGitHubComment(ctx, pr.RepoOwner, pr.RepoName, pr.Number, msg)
+
+	_, _, err = s.GithubClient.Issues.AddLabelsToIssue(ctx, pr.RepoOwner, pr.RepoName, pr.Number, []string{"Lifecycle/1:stale", "Do Not Merge"})
+	if err != nil {
+		msg := fmt.Sprintf("error applying labels %v:\n ```%v```", pr.URL, err.Error())
+		s.logToMattermost(ctx, msg)
```
agnivade, line 66: Just curious: what is the general rule that determines when do we log to mattermost, and when we don't?

server/translations.go
```diff
@@ -25,3 +29,83 @@ func (s *Server) handleTranslationPR(ctx context.Context, pr *model.PullRequest)
...
+	return false
+}
+
+func isOnlyModifyingAllowedFiles(prFiles []*github.CommitFile, languageDir string, languageBaseFile string) bool {
+	for _, prFileName := range getPRFileNamesModifyingLanguageDir(prFiles, languageDir) {
+		if prFileName == languageBaseFile || filepath.Ext(prFileName) == "json" {
+			continue
+		} else {
+			return false
+		}
```
agnivade, line 97: Outdent this. It is not required to be in an else block.

server/translations_test.go
```diff
@@ -0,0 +1,87 @@
+package server
+
+import (
+	"testing"
+
+	"github.com/google/go-github/v32/github"
+	"github.com/mattermost/mattermost-mattermod/model"
+	"github.com/stretchr/testify/assert"
```
agnivade, line 8: Nit: incorrectly grouped imports.

server/translations_test.go
```diff
@@ -0,0 +1,87 @@
...
+)
+
+func TestIsTranslatedRepo(t *testing.T) {
```
agnivade, line 11: TestIsTranslatedRepo and TestIsNotTranslatedRepo can be sub-tests.

server/translations_test.go
```diff
@@ -0,0 +1,87 @@
...
+	translationsRepo := &TranslationsRepo{
+		RepoName: "mattermost",
+	}
+	translationsRepos := []*TranslationsRepo{translationsRepo}
+	translationsRepo, isTranslatedRepo := isTranslatedRepo(pr, translationsRepos)
+	assert.False(t, isTranslatedRepo)
+	assert.Nil(t, translationsRepo)
+}
+
+func TestIsModifyingLanguageDir(t *testing.T) {
```
agnivade, line 33: Same applies for these 4 tests below.

server/translations.go
```diff
@@ -25,3 +29,83 @@ func (s *Server) handleTranslationPR(ctx context.Context, pr *model.PullRequest)
...
+		if t.RepoName == pr.RepoName {
+			return t, true
+		}
+	}
+	return nil, false
+}
+
+func isModifyingLanguageDir(prFiles []*github.CommitFile, languageDir string) bool {
+	for _, prFile := range prFiles {
+		prFilePath, _ := path.Split(prFile.GetFilename())
```
agnivade, line 83: Wondering if `path.Dir` a better choice for this?

server/translations.go
```diff
@@ -25,3 +29,83 @@ func (s *Server) handleTranslationPR(ctx context.Context, pr *model.PullRequest)
...
+		return
+	}
+
+	if s.Config.TranslationsBot == pr.Username {
+		return
+	}
+
+	prFiles, err := s.getPRFiles(ctx, pr.RepoOwner, pr.RepoName, pr.Number)
+	if err != nil {
+		mlog.Error("Error listing PR files", mlog.String("repo", pr.RepoName), mlog.Int("pr", pr.Number), mlog.Err(err))
```
agnivade, line 45: This is okay for this PR. But in future PRs, it's better to propagate the error upwards.

server/translations.go
metanerd: If it is interesting for me to know. In this case, I am interested that these PRs cannot be merged which are created.

agnivade: I see. If that is the sole intention, would it make sense to have Kibana filters set up and monitor log patterns like these? That is the central destinations for all logs and it is easy to set up ad-hoc filters and alerts dynamically without modifying the code.

My point is that monitoring logs should be a separate concern from the code.

Nothing to do right now. Just wanted your opinion on what do you think.

PR #154, opened 2020-07-06
[MM-26580] server/update_branch: add some tests

isacikgoz:
#### Summary
This PR adds some tests to the `update-branch` logic. The first part (getting `model.PullRequest` from the comment) of the handler (`handleUpdateBranch`) is repeated in 3 other handlers, so I skipped testing that part for this PR. In a separate ticket we can reduce the repetition and decouple that logic from this function and others. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26580


server/update_branch.go
```diff
@@ -2,60 +2,84 @@ package server
...
+	"github.com/mattermost/mattermost-mattermod/model"
 
 	"github.com/google/go-github/v32/github"
 )
 
-func (s *Server) handleUpdateBranch(ctx context.Context, eventIssueComment IssueComment) {
+var (
+	ErrCommenterPermission    error = errors.New("commenter does not have permissions")
+	ErrOrganizationPermission error = errors.New("we don't have permissions")
+	ErrUpdatePullRequest      error = errors.New("could not update pull request")
```
agnivade, line 18: I don't think we really need to specify the `error` type. Directly assigning is also fine. Unless you explicitly meant to do that?

server/github.go
```diff
@@ -22,16 +22,17 @@ const (
...
+		RepoName:            *pullRequest.Base.Repo.Name,
+		Number:              *pullRequest.Number,
+		Username:            *pullRequest.User.Login,
+		FullName:            "",
+		Ref:                 *pullRequest.Head.Ref,
+		Sha:                 *pullRequest.Head.SHA,
+		State:               *pullRequest.State,
+		URL:                 *pullRequest.URL,
+		CreatedAt:           pullRequest.GetCreatedAt(),
+		MaintainerCanModify: pullRequest.GetMaintainerCanModify(),
```
agnivade, line 35: Commenting here because of GitHub limitations. But in L80, we are doing the same thing of logging the error. Let's propagate that upwards too.

server/update_branch.go
```diff
@@ -2,60 +2,84 @@ package server
...
 	commenter := eventIssueComment.Comment.User.GetLogin()
+	if err := s.checkUpdatePullRequestPermissions(commenter, pr); err != nil {
+		var msg string
+		switch err {
+		case ErrCommenterPermission:
+			msg = "Looks like you don't have permissions to trigger this command.\n Only available for the PR submitter and org members"
+		case ErrOrganizationPermission:
+			msg = "We don't have permissions to update this PR, please contact the submitter to apply the update."
+		}
+		s.sendGitHubComment(ctx, pr.RepoOwner, pr.RepoName, pr.Number, msg)
```
agnivade, line 45: Amazing refactoring!

server/update_branch.go
```diff
@@ -2,60 +2,84 @@ package server
...
 	}
 
 	commenter := eventIssueComment.Comment.User.GetLogin()
+	if err := s.checkUpdatePullRequestPermissions(commenter, pr); err != nil {
+		var msg string
+		switch err {
+		case ErrCommenterPermission:
+			msg = "Looks like you don't have permissions to trigger this command.\n Only available for the PR submitter and org members"
+		case ErrOrganizationPermission:
+			msg = "We don't have permissions to update this PR, please contact the submitter to apply the update."
```
agnivade, line 43: Let's switch to using `errors.Is` to make it more robust.

server/update_branch.go
```diff
@@ -2,60 +2,84 @@ package server
...
+
 	pr, err := s.GetPullRequestFromGithub(ctx, prGitHub)
 	if err != nil {
-		mlog.Error("Error Updating the PR in the DB", mlog.Err(err))
-		return
+		return fmt.Errorf("error updating the PR in the DB: %w", err)
 	}
 
 	commenter := eventIssueComment.Comment.User.GetLogin()
+	if err := s.checkUpdatePullRequestPermissions(commenter, pr); err != nil {
```
agnivade, line 37: checkUpdatePRPermissions

server/update_branch.go
```diff
@@ -2,60 +2,84 @@ package server
...
+		case ErrCommenterPermission:
+			msg = "Looks like you don't have permissions to trigger this command.\n Only available for the PR submitter and org members"
+		case ErrOrganizationPermission:
+			msg = "We don't have permissions to update this PR, please contact the submitter to apply the update."
+		}
+		s.sendGitHubComment(ctx, pr.RepoOwner, pr.RepoName, pr.Number, msg)
+		return err
+	}
+
+	if err := s.updatePullRequest(ctx, pr); err != nil {
```
agnivade, line 49: I understand what you are trying to do here. By refactoring to a separate function, we can test this in isolation. But I don't think that's really necessary here. I guess the problem was that the store mocks weren't available, so we couldn't really test `s.GetPullRequestFromGithub`.

I think once my PR goes in we can expand our test coverage by testing that part too, and then bringing this method back inside. We can refactor the `s.sendGitHubComment` part using a defer block and just changing the `msg` variable.

server/update_branch_test.go
```diff
@@ -0,0 +1,147 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
agnivade, line 1: Let's wait for store mocks to go in, and add tests for `s.GetPullRequestFromGithub` too.

server/update_branch.go
```diff
@@ -2,60 +2,84 @@ package server
...
 	// If the commenter is not the PR submitter, check if the PR submitter is an org member
 	if commenter != pr.Username && !s.IsOrgMember(commenter) {
-		mlog.Debug("not org member", mlog.String("user", commenter))
-		s.sendGitHubComment(ctx, pr.RepoOwner, pr.RepoName, pr.Number, "Looks like you don't have permissions to trigger this command.\n Only available for the PR submitter and org members")
-		return
+		return ErrCommenterPermission
 	}
 
-	repoInfo := strings.Split(pr.FullName, "/")
+	repoInfo := strings.Split(pr.Username, "/")
```
agnivade, line 63: Was this actually a bug?

server/update_branch_test.go
```diff
@@ -0,0 +1,147 @@
...
+import (
+	"context"
+	"errors"
+	"net/http"
+	"testing"
+
+	"github.com/golang/mock/gomock"
+	"github.com/google/go-github/v32/github"
+	"github.com/mattermost/mattermost-mattermod/model"
+	"github.com/mattermost/mattermost-mattermod/server/mocks"
```
agnivade, line 15: Nit: split out mattermod imports in a separate block. The order should be stdlib, own repo, third party.

server/server.go
```diff
@@ -260,7 +260,8 @@ func (s *Server) githubEvent(w http.ResponseWriter, r *http.Request) {
 			s.handleAutoassign(ctx, *eventIssueComment)
 		}
 		if strings.Contains(strings.TrimSpace(*eventIssueComment.Comment.Body), "/update-branch") {
-			s.handleUpdateBranch(ctx, *eventIssueComment)
+			err := s.handleUpdateBranch(ctx, *eventIssueComment)
+			mlog.Error("Error updating branch", mlog.Err(err))
```
agnivade, line 264: No nil check?

server/update_branch.go
isacikgoz: Actually these can be removed for now. They left from my complete isolation trial :)

server/update_branch.go
isacikgoz: Yes, my motivation was exactly what you've said. Actually this function and the permission check can be bundled together. The remaining part, `s.GetPullRequestFromGithub` should be a separate logic simply because it only converts a `github.PullRequest` to `model.PullRequest` for this functions point of view. And It's repeated several times in other places. If you agree I can include those changes into this PR as well.

server/server.go
isacikgoz: 🤦

server/update_branch.go
isacikgoz: No, I think I mixed up 😬

server/update_branch.go
agnivade: > If you agree I can include those changes into this PR as well.

Up to you. 0/5 on that.

> Actually this function and the permission check can be bundled together. 

By bundled together, you mean in the same method or as a different method? Totally up to you, but I felt we don't really need 2 separate methods for this if we can test everything once we have the store mocks in.

server/update_branch.go
```diff
@@ -2,60 +2,68 @@ package server
...
+	ErrCommenterPermission    = errors.New("commenter does not have permissions")
+	ErrOrganizationPermission = errors.New("we don't have permissions")
+	ErrUpdatePullRequest      = errors.New("could not update pull request")
+)
+
+const (
+	MsgCommenterPermission    = "Looks like you don't have permissions to trigger this command.\n Only available for the PR submitter and org members"
+	MsgOrganizationPermission = "We don't have permissions to update this PR, please contact the submitter to apply the update."
+	MsgUpdatePullRequest      = "Error trying to update the PR.\nPlease do it manually."
+)
```
agnivade, line 25: I have an idea to improve this further. Needs a little bit of more work, but might be more idiomatic. What we are doing here is essentially map the error to the string. So the contents of the error is entirely wasted here, as what gets printed is not the error string, but another message which is mapped to it.

Perhaps what we can do is have error types instead like

```go
type ErrPermission struct {
  source string // this can be commenter/org
}

func (e *ErrPermission) Error() string {
  // Now here we print the proper message depending on the source
}

func newErrPermission(source) *ErrPermission {
 // return the error
}
```

This way we can reduce the duplication in information in the errors and the string messages. Because basically we want to capture errors and then pass the error messages along.

Might be better to do this in a separate PR since a lot of work has already gone in to this one. If you think this is a good idea, let's file a ticket and work on it later.

isacikgoz: Agreed, it would be more convenient to handle errors/messages that way/ There are several occurrences like this in the `server` package that might benefit from this treatment. Created the ticket [here](https://mattermost.atlassian.net/browse/MM-26783) 👍

PR #156, opened 2020-07-07
MM-26453: Add tests for auto_merge.go

agnivade:
This PR creates mocks for the store layer and adds tests for the
AutoMergePR method. We test for 4 functionalities:
- A basic test that checks the happy path to confirm that a PR
in the correct state gets merged and the label removed.
- A PR not having the auto-merge label gets left alone.
- A PR in closed/unclean state gets left alone.
- A PR having pending reviewers gets left alone.

We also tidy up the log lines and change the log level where necessary.

Import groups are also fixed.

Lastly, now the AutoMerge method returns an error instead of logging it.
This makes the API cleaner and easier to test.

#### Ticket link

https://mattermost.atlassian.net/browse/MM-26453


server/auto_merge.go
```diff
@@ -8,68 +8,91 @@ import (
...
-		mlog.Error(err.Error())
-		return
+		return fmt.Errorf("error while listing open PRs %w", err)
 	}
 
 	for _, pr := range prs {
-		if !s.isAutoMergeLabelInLabels(pr.Labels) {
-			mlog.Info("No auto merge label for this PR", mlog.Int("pr", pr.Number), mlog.String("repo", pr.RepoName))
+		if !s.hasAutoMerge(pr.Labels) {
+			mlog.Debug("No auto merge label for this PR; skipping",
```
agnivade, line 28: Is this log line really necessary?

server/auto_merge.go
```diff
@@ -8,68 +8,91 @@ import (
...
+				mlog.String("repo", pr.RepoName),
+				mlog.String("SHAFromPR", ghPR.Head.GetSHA()),
+				mlog.String("SHAFromStatus", prStatus.GetSHA()))
 			continue
 		}
 
-		if PRStatus.GetState() != stateSuccess {
-			mlog.Error("PR is not ready to merge, not ready state", mlog.Int("pr", pr.Number), mlog.String("repo", pr.RepoName), mlog.String("state", PRStatus.GetState()))
+		if prStatus.GetState() != stateSuccess {
+			mlog.Debug("PR is not ready to merge; not in success state",
```
agnivade, line 75: Same as above.

agnivade:
@metanerd - I have 2 questions above. Do you think these log lines are necessary?

server/auto_merge.go
metanerd: No, not necessary.

server/auto_merge.go
metanerd: Same as above.

agnivade:
I think only "// Get the Statuses" is obvious enough to be removed. For the other two, the `ListReviewers` call isn't immediately apparent that it only lists pending reviewers. I got confused for a while there. And `// All good to merge` isn't too useless IMO. It indicates that all the conditions are met, and now we are merging it.

I can drop the "// Get the Statuses" if you think so.

PR #157, opened 2020-07-07
MM-26381: remove AppError from codebase

agnivade:
The mattermod codebases still uses AppError which is just an artifact
from the main server codebase. There is no actual need for an AppError, because:

- There is no App.
- Mattermod is not an HTTP API server to pass HTTP responses.

While here, we also remove the FromJSON/ToJSON methods/functions which weren't used at all.

### Ticket link

https://mattermost.atlassian.net/browse/MM-26381


agnivade:
I think that was the fastest review of all time.

PR #158, opened 2020-07-07
[MM-26578] store/sql_pull_request_store: add tests

isacikgoz:
#### Summary
This PR adds tests to the `store/sql_pull_request_store` ~with using [go-sqlmock](https://github.com/DATA-DOG/go-sqlmock)~.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26578

### DIFF
--- a/store/sql_pull_request_store_test.go
+++ b/store/sql_pull_request_store_test.go
@@ -11,8 +11,7 @@ import (
 )
 
 func TestPullRequestStore(t *testing.T) {
-	ss, teardown := getTestSQLStore(t)
-	defer teardown()
+	ss := getTestSQLStore(t)
 
 	prs := NewSQLPullRequestStore(ss)
 	prs.(*SQLPullRequestStore).CreateIndexesIfNotExists()
--- a/store/sql_pull_request_store_test.go
+++ b/store/sql_pull_request_store_test.go
@@ -1,0 +1,107 @@
+package store
+
+import (
+	"errors"
+	"regexp"
+	"testing"
+
+	"github.com/mattermost/mattermost-mattermod/model"
+
+	"github.com/DATA-DOG/go-sqlmock"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestPullRequestStore(t *testing.T) {
+	ss, mock, teardown := getMockSQLStore(t)
+	defer teardown()
+
+	prs := NewSQLPullRequestStore(ss)
+
+	t.Run("happy path on Save", func(t *testing.T) {
+		pr := &model.PullRequest{
+			RepoOwner: "owner",
+			RepoName:  "repo-name",
+			Number:    123,
+		}
+
+		mock.ExpectExec("update `PullRequests`").WillReturnResult(sqlmock.NewResult(1, 1))
+
+		npr, err := prs.Save(pr)
+		require.NoError(t, err)
+		assert.Equal(t, pr.RepoOwner, npr.RepoOwner)
+		assert.Equal(t, pr.RepoName, npr.RepoName)
+		assert.Equal(t, pr.Number, npr.Number)
+	})
+
+	t.Run("err on Save", func(t *testing.T) {
+		pr := &model.PullRequest{}
+
+		mock.ExpectExec("update `PullRequests`").WillReturnError(errors.New("some-error"))
+
+		npr, err := prs.Save(pr)
+		require.Error(t, err)
+		assert.Nil(t, npr)
+	})
+
+	t.Run("happy path Get", func(t *testing.T) {
+		pr := &model.PullRequest{
+			RepoOwner: "owner",
+			RepoName:  "repo-name",
+			Number:    123,
+		}
+
+		rows := sqlmock.NewRows([]string{"RepoOwner", "RepoName", "Number"}).
+			AddRow("owner", "repo-name", 123)
+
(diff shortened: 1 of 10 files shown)


agnivade:
Sorry I should have been clearer with this. I would prefer proper integration tests. We want to actually test the DB here. It can be mocked for other packages, but here we want to see that the data is stored properly in the DB.

store/sql_store_test.go
```diff
@@ -0,0 +1,110 @@
...
+	defaultMysqlDB      = "mattermod_test"
+)
+
+type teardownFunc func()
+
+func getTestSQLStore(t *testing.T) (*SQLStore, teardownFunc) {
+	teardownSuite, err := createTempDB(defaultMysqlDB, getEnv("MYSQL_USER", defaultMysqlUser))
+	if err != nil {
+		t.Fatal(err)
+	}
```
hanzei, line 28: Objections on using testify?

store/sql_store_test.go
```diff
@@ -0,0 +1,110 @@
...
+	defaultMysqlUser    = "mattermod"
+	defaultMysqlRootPWD = "mattermod"
+	defaultMysqlUserPWD = "mattermod"
+	defaultMysqlDB      = "mattermod_test"
+)
+
+type teardownFunc func()
+
+func getTestSQLStore(t *testing.T) (*SQLStore, teardownFunc) {
+	teardownSuite, err := createTempDB(defaultMysqlDB, getEnv("MYSQL_USER", defaultMysqlUser))
```
hanzei, line 25: Is this a [`t.Helper()`](https://golang.org/pkg/testing/#T.Helper)?

store/sql_store_test.go
```diff
@@ -0,0 +1,110 @@
...
+	defaultMysqlDSN     = "mattermod:mattermod@tcp(localhost:3306)/mattermost_mattermod_test?charset=utf8mb4,utf8\u0026readTimeout=30s\u0026writeTimeout=30s\u0026parseTime=true"
+	defaultMysqlUser    = "mattermod"
+	defaultMysqlRootPWD = "mattermod"
+	defaultMysqlUserPWD = "mattermod"
+	defaultMysqlDB      = "mattermod_test"
+)
+
+type teardownFunc func()
+
+func getTestSQLStore(t *testing.T) (*SQLStore, teardownFunc) {
```
hanzei, line 24: Can we make use of [`t.Cleanup`](https://golang.org/pkg/testing/#T.Cleanup)  here instead of returning a `func()`?

store/sql_store_test.go
```diff
@@ -0,0 +1,110 @@
...
+	if err := store.master.CreateTablesIfNotExists(); err != nil {
+		teardown()
+		t.Fatal(err)
+	}
+
+	return store, teardown
+}
+
+func createTempDB(dbName, dbUser string) (teardownFunc, error) {
+	rootPwd := getEnv("MYSQL_ROOT_PASSWORD", defaultMysqlRootPWD)
```
hanzei, line 71: Similar questions as above

store/sql_store_test.go
isacikgoz: Definitely! Thanks @hanzei 👍

store/sql_store_test.go
isacikgoz: I think it's not needed here, we don't check for specific errors or specific outputs. We just don't want any errors on the path. Also, I may missing something.

hanzei: `testify` has become the standard way of asserting any output from a test. I personally quite like the way the output looks like and therefore prefer to use it everywere.

store/sql_pull_request_store_test.go
```diff
@@ -0,0 +1,53 @@
...
+	})
+
+	t.Run("happy path Get", func(t *testing.T) {
+		npr, err := prs.Get(pr.RepoOwner, pr.RepoName, pr.Number)
+		require.NoError(t, err)
+		assert.NotNil(t, npr)
+		assert.Equal(t, npr.RepoOwner, pr.RepoOwner)
+	})
+
+	t.Run("happy path on ListOpen", func(t *testing.T) {
```
agnivade, line 45: Let's save the same PR with state closed, and then run a listOpen again to verify that no rows are returned.

store/sql_store_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+	db, err := sql.Open("mysql", cfg.FormatDSN())
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	store := &SQLStore{
+		master: &gorp.DbMap{
+			Db:            db,
+			TypeConverter: mattermConverter{},
```
agnivade, line 45: I guess then https://mattermost.atlassian.net/browse/MM-26710 is not required anymore?

store/sql_store_test.go
```diff
@@ -0,0 +1,102 @@
...
+			Dialect: gorp.MySQLDialect{
+				Engine:   "InnoDB",
+				Encoding: "UTF8MB4",
+			},
+		},
+	}
+	t.Cleanup(func() {
+		db.Close()
```
agnivade, line 53: check error here?

store/sql_store_test.go
isacikgoz: Yes it looks like the custom scanner is required for now.

store/sql_store_test.go
isacikgoz: I see, thanks for the explanation. Let's continue with `testing` and once we complete the tests in the `store` package. We can spike and convert these to use `testify` instead.

PR #160, opened 2020-07-08
Add missing colmap

ethervoid:
#### Summary

This PR adds a missing colmap not included in #154


ethervoid:
I'm also going to include the needed migration here. I'm going to create a new ticket to include a DB migration process in mattermod too

agnivade:
> I'm going to create a new ticket to include a DB migration process in mattermod too

That sounds great. Let's chalk up a proper migration design from scratch and think of if we can remove gorp altogether. These will come handy while applying the learnings to the server repo.

PR #164, opened 2020-07-08
[MM-26798] Fix automerge error with null columns

ethervoid:
#### Summary

This PR:

- Fixes the automerge error with null values from the new column called `MaintainerCanModify`
- Includes the used migration SQL to upgrade the DB including the new column

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-26798


ethervoid:
@metanerd When this changed is deployed is going to update a ton of PRs because is going to update the `maintainer_can_modify` flag for all of them. This only is going to affect open PRs given that the `Tick` method only updates those

model/pull_request.go
```diff
@@ -27,5 +28,5 @@ type PullRequest struct {
 	BuildLink           string
 	URL                 string
 	CreatedAt           time.Time
-	MaintainerCanModify bool
+	MaintainerCanModify sql.NullBool
```
agnivade, line 31: Hmm.. this is okay for now, but I'll prefer a separate mapping which maps SQL objects to model objects. Right now, this leaks the abstraction that `PullRequest` is tied to a SQL query which does not make it very flexible.

We can scan a separate object having `sql.NullBool` and then map it to PullRequest which can have a normal `bool` pointer field.

ethervoid: Agree

PR #166, opened 2020-07-09
MM-26797: Fix context cancellation on goroutines

agnivade:
We passed several contexts to goroutines. Therefore when the parent
function returned, the context got cancelled and therefore the running
goroutines would throw an error. To fix this:

- We remove passing the context in checkIfNeedCherryPick. This is not
ideal and every call should take the parent context. The problem is that
a request handler is semi-synchronous. It does some jobs synchronously and
spins of time taking jobs in goroutines without any control of how to shut
them down.

What we need is to make the entire event handler async, and create a context
inside that and run everything synchronously inside that. Until we have that,
we need to make do with this.

- We make createCLAPendingStatus synchronous. This also fixes a long-standing
bug where the CLA check would randomly go to a pending state.

- We move sendGitHubComment to a closure and create a new context. Again, not
ideal but have to bear with this.

#### Ticket link

https://mattermost.atlassian.net/browse/MM-26797

### DIFF
--- a/server/cherry_pick.go
+++ b/server/cherry_pick.go
@@ -55,7 +55,10 @@ func (s *Server) handleCherryPick(ctx context.Context, eventIssueComment IssueCo
 	}
 }
 
-func (s *Server) checkIfNeedCherryPick(ctx context.Context, pr *model.PullRequest) {
+func (s *Server) checkIfNeedCherryPick(pr *model.PullRequest) {
+	ctx, cancel := context.WithTimeout(context.Background(), defaultRequestTimeout*time.Second)
+	defer cancel()
+
 	prCherryCandidate, _, err := s.GithubClient.PullRequests.Get(ctx, pr.RepoOwner, pr.RepoName, pr.Number)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


server/cherry_pick.go
```diff
@@ -55,7 +55,10 @@ func (s *Server) handleCherryPick(ctx context.Context, eventIssueComment IssueCo
 	}
 }
 
-func (s *Server) checkIfNeedCherryPick(ctx context.Context, pr *model.PullRequest) {
+func (s *Server) checkIfNeedCherryPick(pr *model.PullRequest) {
+	ctx, cancel := context.WithTimeout(context.Background(), defaultRequestTimeout*time.Second)
```
isacikgoz, line 59: Maybe a couple of sentences that explains why we are creating context here would be useful for future refactors.

ethervoid:
Yep, I've been thinking the same and I know some special cases would arise because we're mixing sync with async.

agnivade:
@metanerd - Needs another deploy. This fixes cherry picks.

PR #167, opened 2020-07-09
MM-26829: Fix duplicate label removal

agnivade:
There's no need to remove a label after doing auto-merge, because
it already gets removed as part of a cleanup label process when a PR
is closed.

This fixes a race condition where we were trying to remove the same label twice,
generating logs like this:

```
ERRO[2020-07-09T18:54:14.659247+05:30] Error removing the label                      caller="server/github.go:129" error="DELETE https://api.github.com/repos/mattermost/mattermost-server/issues/14970/labels/AutoMerge: 404 Label does not exist []
```

While here, we add a continue statement to quickly move forward.

### Ticket link

https://mattermost.atlassian.net/browse/MM-26820


agnivade:
@metanerd - needs a deploy.

metanerd:
Done

PR #168, opened 2020-07-10
[MM-26841] server: gently refactor github handler v1

isacikgoz:
#### Summary
This PR hpefully fixes the magical activity in the `handleAutoAssign` method. Tried to refactor the GitHub event handler function to be more readable and more reliable.

PS: We always need a PR issue number to be successful in `handleAutoAssign` see the old code [here](https://github.com/mattermost/mattermost-mattermod/blob/8d1283949607f2ed7abe2f5a09c833d389a2f57b/server/autoassign.go#L34).

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26841

### DIFF
--- a/server/github_pr.go
+++ b/server/github_pr.go
@@ -28,32 +28,32 @@ type EventData struct {
 	Repository *github.Repository         `json:"repository"`
 }
 
-// CheckCLA is true if body contains "/check-cla"
-func (d *EventData) CheckCLA() bool {
+// HasCheckCLA is true if body contains "/check-cla"
+func (d *EventData) HasCheckCLA() bool {
 	if d.Comment == nil || d.Comment.Body == nil {
 		return false
 	}
 	return strings.Contains(strings.TrimSpace(*d.Comment.Body), "/check-cla")
 }
 
-// CherryPick is true if body contains "/cherry-pick"
-func (d *EventData) CherryPick() bool {
+// HasCherryPick is true if body contains "/cherry-pick"
+func (d *EventData) HasCherryPick() bool {
 	if d.Comment == nil || d.Comment.Body == nil {
 		return false
 	}
 	return strings.Contains(strings.TrimSpace(*d.Comment.Body), "/cherry-pick")
 }
 
-// AutoAssign is true if body contains "/autoassign"
-func (d *EventData) AutoAssign() bool {
+// HasAutoAssign is true if body contains "/autoassign"
+func (d *EventData) HasAutoAssign() bool {
 	if d.Comment == nil || d.Comment.Body == nil {
 		return false
 	}
 	return strings.Contains(strings.TrimSpace(*d.Comment.Body), "/autoassign")
 }
 
-// UpdateBranch is true if body contains "/update-branch"
-func (d *EventData) UpdateBranch() bool {
+// HasUpdateBranch is true if body contains "/update-branch"
+func (d *EventData) HasUpdateBranch() bool {
 	if d.Comment == nil || d.Comment.Body == nil {
 		return false
 	}
--- a/server/github_pr.go
+++ b/server/github_pr.go
@@ -6,6 +6,7 @@ package server
 import (
 	"encoding/json"
 	"io"
+	"strings"
 
 	"github.com/google/go-github/v32/github"
 )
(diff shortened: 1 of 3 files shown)


server/github_pr.go
```diff
@@ -20,13 +21,45 @@ type PullRequestEvent struct {
...
 	Issue      *github.Issue              `json:"issue"`
 	Repository *github.Repository         `json:"repository"`
 }
 
+// CheckCLA is true if body contains "/check-cla"
+func (d *EventData) CheckCLA() bool {
+	if d.Comment == nil || d.Comment.Body == nil {
+		return false
+	}
+	return strings.Contains(strings.TrimSpace(*d.Comment.Body), "/check-cla")
```
ethervoid, line 36: What do you think of using regexp to be sure using something like `\b(/check-cla)\b`? 2/5

server/pull_request.go
```diff
@@ -519,11 +520,15 @@ func (s *Server) isBlockPRMergeInLabels(labels []string) bool {
 	return false
 }
 
-func (s *Server) getPRFromComment(ctx context.Context, comment IssueComment) (*model.PullRequest, error) {
+func (s *Server) getPRFromEvent(ctx context.Context, event EventData) (*model.PullRequest, error) {
+	if event.Issue == nil || event.Repository == nil {
+		return nil, errors.New("invalid event")
```
ethervoid, line 525: I'd include a more meaningful error. Why is the event invalid?

server/server.go
```diff
@@ -223,60 +223,59 @@ func (s *Server) githubEvent(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	var pingEvent *github.PingEvent
 	if r.Header.Get("X-GitHub-Event") == "ping" {
-		pingEvent = PingEventFromJSON(ioutil.NopCloser(bytes.NewBuffer(buf)))
+		if pingEvent := PingEventFromJSON(ioutil.NopCloser(bytes.NewBuffer(buf))); pingEvent != nil {
+			mlog.Info("ping event", mlog.Int64("HookID", pingEvent.GetHookID()))
```
ethervoid, line 228: Info o Warn level?

server/server.go
```diff
@@ -223,60 +223,59 @@ func (s *Server) githubEvent(w http.ResponseWriter, r *http.Request) {
...
+	if eventData.Comment != nil && eventData.Comment.User != nil {
+		commenter = eventData.Comment.User.GetLogin()
 	}
 
-	if eventIssueComment != nil && eventIssueComment.Action == "created" {
-		if strings.Contains(strings.TrimSpace(*eventIssueComment.Comment.Body), "/check-cla") {
-			if err := s.handleCheckCLA(ctx, pr); err != nil {
-				mlog.Error("Error checking CLA", mlog.Err(err))
-			}
+	if eventData.CheckCLA() {
```
ethervoid, line 256: Maybe we could change all this CHeck explosion into a factory that returns the proper handler to be executed. What do you think? The rationale behind is that as we would increase then umber of commands we would have a ton more of this boilerplate

agnivade:
Should we still work on this if the handlers will be split to different endpoints? Wondering if it makes sense to start with a fresh approach migrating each event one by one.

isacikgoz:
I'd like to start fresh TBH. But we need to coordinate with admins to also update hooks on GH configuration side. This PR specifically does not make a huge impact just a housekeeping related to naming and function flow.

server/github_pr.go
isacikgoz: Don't have any strong opinions, I tend to use regexp on more complex stuff :)

server/server.go
isacikgoz: This was `info` level before, so, if you'd ask my opinion; I'd make this `trace` level :)

server/server.go
isacikgoz: Once we have a better function signature on handlers, we can definitely go to that path. I was thinking the same with you actually. 👍

ethervoid:
I'd perform this refactor to make it more approachable and open a new issue to make those deeper changes

agnivade:
Sounds good. But won't we have to rename `EventData` to `IssueComment` again? I have no problems with it, just checking if this refactor doesn't become useless if we move to the new model.

Secondly, let's wait until the revert PR which I sent gets merged? There might be merge conflicts then.

ethervoid:
> But won't we have to rename EventData to IssueComment again? I have no problems with it, just checking if this refactor doesn't become useless if we move to the new model.

I think right now makes sense. Given that the work is done I'll merge it and if we have to change it then we'll do it

> Secondly, let's wait until the revert PR which I sent gets merged? There might be merge conflicts then.

What PR are we talking about?

agnivade:
Ok, sounds good. The PR was https://github.com/mattermost/mattermost-mattermod/pull/170. I guess everything is sorted now.

server/github_pr.go
```diff
@@ -20,13 +21,45 @@ type PullRequestEvent struct {
...
 
-type IssueComment struct {
+type EventData struct {
 	Action     string                     `json:"action"`
 	Comment    *github.PullRequestComment `json:"comment"`
 	Issue      *github.Issue              `json:"issue"`
 	Repository *github.Repository         `json:"repository"`
 }
 
+// CheckCLA is true if body contains "/check-cla"
```
agnivade, line 31: I would rename these methods to have a "Has" prefix to be more clear. Right now. `e.CheckCLA` indicates this method is checking CLA which is not right.

Something like HasCheckCLA/HasCherryPick/HasAutoAssign/HasUpdateBranch

server/github_pr.go
agnivade: A simple `strings.Contains` is fine IMO.

server/server.go
```diff
@@ -223,60 +223,59 @@ func (s *Server) githubEvent(w http.ResponseWriter, r *http.Request) {
...
-
 	if event != nil && event.PRNumber != 0 {
 		mlog.Info("pr event", mlog.Int("pr", event.PRNumber), mlog.String("action", event.Action))
 		s.handlePullRequestEvent(ctx, event)
 		return
 	}
 
-	pr, err := s.getPRFromComment(ctx, *eventIssueComment)
+	eventData := EventDataFromJSON(ioutil.NopCloser(bytes.NewBuffer(buf)))
+	if eventData == nil || eventData.Action != "created" {
```
agnivade, line 241: This is confusing :) We are checking for something else, and passing something else in the method. Can't wait to refactor this.

To be extra sure, can we do some manual testing of the 3 cases (handlePullRequestEvent/handleIssueEvent/rest) by constructing some curl requests and checking that the correct code paths are being followed? This looks okay from code, but without any actual tests, I am not really confident.

isacikgoz: Agree with you on the confusion, this will hopefully be lost in dusty pages of history pretty soon. As for the `curl` requests, with simple data we can confirm high level paths but it will going to fail (panic) eventually since we need a real data to continue on communication with GitHub. Otherwise I may need some real GH requests to test these.

agnivade: > but it will going to fail (panic) eventually since we need a real data to continue on communication with GitHub. 

Why would it panic? I would assume that the request will fail because it's not authenticated.

isacikgoz: You're right, it would fail before it panics. So this needs to be tested in a staging env.?

agnivade: There's no staging env yet. I was thinking of just manually testing with curl. The request can fail eventually. But I wanted to verify that the right code paths are being taken.

PR #169, opened 2020-07-10
[MM-26454] Add tests for block_pr_merge

ethervoid:
#### Summary

This PR includes:

- Tests for the `block_pr_merge` code
- Performs a little refactor to make easier to test some cases

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-26454

### DIFF
--- a/server/block_pr_merge.go
+++ b/server/block_pr_merge.go
@@ -12,9 +12,9 @@ import (
 	"github.com/mattermost/mattermost-server/v5/mlog"
 )
 
-func (s *Server) blockPRMerge(ctx context.Context, pr *model.PullRequest) {
+func (s *Server) blockPRMerge(ctx context.Context, pr *model.PullRequest) error {
 	if pr.State == model.StateClosed {
-		return
+		return nil
 	}
 
 	mergeStatus := &github.RepoStatus{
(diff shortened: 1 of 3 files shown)


server/block_pr_merge.go
```diff
@@ -26,10 +26,7 @@ func (s *Server) blockPRMerge(ctx context.Context, pr *model.PullRequest) {
 
 	mlog.Info("will block PR merge status", mlog.Int("pr", pr.Number), mlog.String("repo", pr.RepoName))
 	_, _, errStatus := s.GithubClient.Repositories.CreateStatus(ctx, pr.RepoOwner, pr.RepoName, pr.Sha, mergeStatus)
```
agnivade, line 28: Might as well rename this to `err`

server/block_pr_merge.go
```diff
@@ -57,8 +54,5 @@ func (s *Server) unblockPRMerge(ctx context.Context, pr *model.PullRequest) {
 
 	mlog.Info("will unblock PR merge status", mlog.Int("pr", pr.Number), mlog.String("repo", pr.RepoName))
 	_, _, errStatus := s.GithubClient.Repositories.CreateStatus(ctx, pr.RepoOwner, pr.RepoName, pr.Sha, mergeStatus)
```
agnivade, line 56: Same as above.

server/pull_request.go
```diff
@@ -535,3 +527,15 @@ func (s *Server) getPRFromComment(ctx context.Context, comment IssueComment) (*m
 	}
 	return pr, nil
 }
+
+func (s *Server) setBlockStatusForPR(ctx context.Context, pr *model.PullRequest) {
```
isacikgoz, line 531: shall we propagate error here as well?

ethervoid: I thought about it but if we propagate the error then we're going to have multiple `mlog.Error` with the same message because that error is not returning a 4xx error by design so if the outcome of that error is showing an error trace then should be better be encapsulated inside that method

PR #171, opened 2020-07-13
MM-26756: cherry-pick instead of am

ashishbhate:
#### Summary

- git cherry-pick instead of git am
#### Ticket Link

- https://mattermost.atlassian.net/browse/MM-26756


ashishbhate:
@ethervoid @agnivade @metanerd 

The github api does not support cherry picking. There are some stackoverflow answers about using a mix of ref updates and merges to emulate cherry-picking using the Github API. I don't think that's the right way to go about it.

go-git doesn't support cherry-picking https://github.com/go-git/go-git/blob/master/COMPATIBILITY.md

We could use [git2go](https://github.com/libgit2/git2go), which is a cgo wrapper around libgit2.

In this PR I've patched and simplified the `cherry-pick.sh` script to actually use `git cherry-pick` instead of `git am`. This isn't a great solution either. Its not easily testable.

What do you guys think?

ashishbhate:
This is PR is not ready for review yet. I've opened it to solicit feedback and get some discussion going about cherry-picking.

agnivade:
Sounds fine to me.

Couple of thoughts on adding the new field: do we need the field to be present in the DB? If not, could we just do `db:"-"`, and skip the whole process of adding a new column? I think we might have inadvertently added some new columns when they aren't strictly required.

ashishbhate:
>Couple of thoughts on adding the new field: do we need the field to be present in the DB? If not, could we just do db:"-", and skip the whole process of adding a new column? I think we might have inadvertently added some new columns when they aren't strictly required.

The field does not need to be in the DB. Everything was being added to the DB so I added this field as well.
I assumed that (eventually) the DB was supposed to be a cache for the Github API (since we were hitting API rate limits), so adding all the columns seemed to make sense. What's the roadmap for the DB stuff?

agnivade:
See my doc here for the long term plan: https://docs.google.com/document/d/1dljwOY5esvJmffKbNbpcdSnM83M-6adXsINvHI_9nEY/edit#. The initial caching client that we put in is working pretty nicely for the moment (although we are adding metrics right now to get a better picture). In the long term, we should move to using maintner as a backend to store the data.

So the DB is going to go away _eventually_. It can even be done away with if we see the caching client to be very successful. I would prefer not to add more fields to it unless it is needed IMO.

ashishbhate:
Looks good. I'll drop the column addition.

ashishbhate:
@agnivade @ethervoid @metanerd 

Ready for review.

ashishbhate:
> While we are here, can we move all the functions at the top and then start the code in a linear fashion, and not have function declarations and calls interleaved?

Do you mean the bash code?

>How have you tested this? :)

I haven't tested the Go stuff end to end. I tested bits and pieces of it by writing a quick scripts that would call code that I changed without having to set up the whole app. 

The bash script I tested manually by running it directly. 
For conflicts checked the error on the terminal (which is captured by the go code and displayed on the original PR)
For successful cherry-picks I made a test repo and had the script run against that repo. https://github.com/ashishbhate/test-repo/pulls

Certainly nothing comprehensive.

agnivade:
> Do you mean the bash code?

Yep.

> The bash script I tested manually by running it directly. I made a test repo and intentionally created conflicts:
For conflicts I checked the error output on the terminal (which is captured by the Go code and displayed in the original PR)
For successful cherry-picks you can see the output at https://github.com/ashishbhate/test-repo/pulls

Fantastic. We can't really test e2e until we have the staging env ready. I just wanted to manually test the script. Thanks for all that work.

Once you re-arrange the bash, lgtm from my side.

hack/cherry-pick.sh
```diff
@@ -105,12 +102,12 @@ echo "+++ Creating local branch ${NEWBRANCHUNIQ}"
 
 cleanbranch=""
 prtext=""
-gitamcleanup=false
-function return_to_kansas {
```
ethervoid, line 109: :laughing:

hack/cherry-pick.sh
```diff
@@ -126,9 +123,8 @@ function return_to_kansas {
     fi
   fi
 }
-trap return_to_kansas EXIT
+trap do_cleanup EXIT
```
ethervoid, line 126: One question. `do_cleanup` relies on the `gitcleanup` variable which is set to false and after this called is set to true but this function is never called again. My question is, are we calling that method to clean something up?

model/pull_request.go
```diff
@@ -29,5 +29,6 @@ type PullRequest struct {
 	URL                 string
 	CreatedAt           time.Time
 	Merged              sql.NullBool
+	MergeCommitSHA      string `db:"-"`
```
ethervoid, line 32: Did you check what happen with the current SQL query if the CommitSHA is null?

hack/cherry-pick.sh
```diff
@@ -50,12 +53,11 @@ if ! which hub > /dev/null; then
...
 if [[ "$#" -lt 2 ]]; then
-  echo "${0} <remote branch> <pr-number>...: cherry pick one or more <pr> onto <remote branch> and leave instructions for proposing pull request"
+  echo "${0} <remote branch> <pr-number> <commit-sha>: cherry pick a <commit> from <pr> onto <remote branch> and leave instructions for proposing pull request"
   echo
-  echo "  Checks out <remote branch> and handles the cherry-pick of <pr> (possibly multiple) for you."
+  echo "  Checks out <remote branch> and handles the cherry-pick of <commits> for you."
   echo "  Examples:"
-  echo "    $0 upstream/release-3.14 12345        # Cherry-picks PR 12345 onto upstream/release-3.14 and proposes that as a PR."
-  echo "    $0 upstream/release-3.14 12345 56789  # Cherry-picks PR 12345, then 56789 and proposes the combination as a single PR."
+  echo "    $0 upstream/release-3.14 12345 df243fd # Cherry-picks commit sha df243fd onto upstream/release-3.14 and proposes that as a PR with a commit message referring to PR 12345"
```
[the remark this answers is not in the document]
ashishbhate: We never used the functionality of merging two PRs and proposing it as a single PR. We only used the single PR "cherry-pick"

I've simplified the code to do a real cherry-pick of a single commit. We can add support for cherry-picking multiple commits, but out current feature branch cherry-pick workflow doesn't require it at the moment.

hack/cherry-pick.sh
```diff
@@ -71,20 +73,15 @@ if git_status=$(git status --porcelain --untracked=no 2>/dev/null) && [[ -n "${g
   exit 1
 fi
 
-if [[ -e "${REBASEMAGIC}" ]]; then
-  echo "!!! 'git rebase' or 'git am' in progress. Clean up and try again."
+while unmerged=$(git status --porcelain | grep ^U) && [[ -n ${unmerged} ]] || [[ -e "${REBASEMAGIC}" ]]; do
```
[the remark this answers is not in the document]
ashishbhate: >Why use while instead of if?

good catch, I'll update to use if

>What doest REBASEMAGIC means or comes from?

Defined on line 34. Its a way to check if a rebase is in process. I'm not sure if our local checkout could ever be in such a state (maybe if someone modified the local checkout on disk directly), but the original code used it to check if the local checkout was in a consistent state and I think its a good idea to continue using it.

hack/cherry-pick.sh
ashishbhate: trap is like defer in go. The function is called when the script exits.

model/pull_request.go
ashishbhate: I dont understand, this field is not added to the DB. 

The cherry-pick code is called either when a PR is closed or the label is added. Both times the PR object is taken from the request, not from the DB. 

Before we do a cherry-pick we check if `MergeCommitSHA` is not empty. If its empty we return error.

hack/cherry-pick.sh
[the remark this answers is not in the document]
ethervoid: Yeah, just wanted to be sure :)

hack/cherry-pick.sh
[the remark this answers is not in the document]
ethervoid: Oh I searched for it and didn’t find it

hack/cherry-pick.sh
ethervoid: Oh good to know I thought it worked differently :)

model/pull_request.go
ethervoid: I’m referring to just double check if the query tries to scan that field and fails. Vanilla gorp should avoid that field and our branch will not. It was just to be sure before deploy the change

hack/cherry-pick.sh
```diff
@@ -95,7 +92,7 @@ if ! git log -n1 --format=%H "${BRANCH}" >/dev/null 2>&1; then
   exit 1
 fi
 
-NEWBRANCHREQ="automated-cherry-pick-of-${PULLDASH}" # "Required" portion for tools.
+NEWBRANCHREQ="automated-cherry-pick-of-#${PULL}" # "Required" portion for tools.
```
metanerd, line 95: We have some git branching logic for building server/webapp/enterprise repo together. The naming of the cherry-pick branches could collide and there is also no cleanup in the repository of the cherry-picking bot, so it could happen that outdated server branch with the equally named webapp cherry pick pull request collides and then it is confusing why the pipeline fails. one solution would be to add somehow the project name into the branch name. Maybe as an acronym.

ashishbhate: I've added the repo name to the branch

ashishbhate:
@agnivade @metanerd @ethervoid 

I've updated the PR. I've been running a local instance of mattermod on my machine linked to https://github.com/ashish-test-org/test-repo you can look at the PRs there for successful and failed cherry-picks.

model/pull_request.go
ashishbhate: @ethervoid I have a mattermod instance running locally and the DB accesses look good. I haven't seen an error from grop for the skipped field

PR #174, opened 2020-07-14
[MM-26723] Metrics system

ethervoid:
#### Summary

This PR adds a metrics system to mattermod. It includes:

- Metrics server
- Prometheus integration
- New metrics transport that gets metrics from GitHub calls
- Metrics for the cron tasks (duration and errors)
- Metrics for the incoming requests and webhook counter by type

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-26723


ethervoid:
@isacikgoz @agnivade please take a look at this first version. I've tried to improve the system we have in the server and I want to know your opinion

agnivade:
Great stuff! Interfaces ending with "er" :) You've made me happy.

I would just suggest some small naming changes to make the API even more beautiful.

- metrics.NewServer
- (*Server).Start
- (*Server).Stop

agnivade:
Nice. Any screenshots of some Prom dashboards that we can look at? :)

ethervoid:
@agnivade Not yet because I can't launch in my local environment and we don't have staging yet. So my plan is to deploy this, once the review passed, and then start working on the dashboard :D

cmd/mattermost-mattermod/main.go
```diff
@@ -27,8 +28,14 @@ func main() {
 	}
 	server.SetupLogging(config)
 
+	// Metrics system
+	metricsProvider := metrics.NewPrometheusProvider()
+	metricsServer := metrics.NewServer("8067", metricsProvider.Handler(), true)
+	metricsServer.Start()
+	defer metricsServer.Stop()
```
isacikgoz, line 35: Should this started and terminated by `server`?

metrics/provider.go
```diff
@@ -0,0 +1,196 @@
...
+	cronNamespace    = "cron"
+	githubNamespace  = "github"
+
+	defaultPrometheusTimeoutSeconds = 60
+)
+
+// Provider is the interface that exposes the communication with the metrics system
+// this interface should be implemented by the different providers we want to include
+type Provider interface {
+	// ObserverHTTPRequestDuration stores the elapsed time for an HTTP request
```
isacikgoz, line 25: Should we use RTT instead?
```suggestion
	// ObserverHTTPRequestDuration stores the round-trip time for an HTTP request
```

metrics/provider.go
```diff
@@ -0,0 +1,196 @@
...
+	httpNamespace    = "requests"
+	cronNamespace    = "cron"
+	githubNamespace  = "github"
+
+	defaultPrometheusTimeoutSeconds = 60
+)
+
+// Provider is the interface that exposes the communication with the metrics system
+// this interface should be implemented by the different providers we want to include
+type Provider interface {
```
isacikgoz, line 24: Shall we rename this into something like `MetricProvider` and move into `server` package? We are losing the interface advatages by coupling this package with server package.

metrics/provider.go
```diff
@@ -0,0 +1,196 @@
...
+
+// PrometheusProvider is the implementation of the Provider interface
+// to send metrics to Prometheus
+type PrometheusProvider struct {
+	Registry *prometheus.Registry
+
+	httpRequestsDuration *prometheus.HistogramVec
+	webhookEvents        *prometheus.CounterVec
+
+	cronTasksDuration *prometheus.HistogramVec
```
isacikgoz, line 54: Is this metric going to be meaningful if we collect for all of the cron jobs. AFAIK we have several cron jobs? I mean we should observe these duration per cron job type.

metrics/server.go
```diff
@@ -0,0 +1,116 @@
...
+	m.server = &http.Server{
+		Addr:         fmt.Sprintf(":%s", m.port),
+		Handler:      router,
+		ReadTimeout:  time.Duration(defaultHTTPServerReadTimeoutSeconds) * time.Second,
+		WriteTimeout: time.Duration(defaultHTTPServerWriteTimeoutSeconds) * time.Second,
+	}
+
+	go func() {
+		mlog.Info("Metrics and profiling server started", mlog.String("port", "8067"))
+		if err := m.server.ListenAndServe(); err != nil {
```
isacikgoz, line 70: Probably I'm missing something but could we include these endpoints to mattermod server?

cmd/mattermost-mattermod/main.go
ethervoid: I want to decouple the metrics system from the server system taking advantage that we are using it for the cron tasks too. So I think is better to start/stop it outside the server cycle

metrics/provider.go
ethervoid: Well, the provider is part of the metrics system and is not related only to the server package. Why are we losing interface advantages?

metrics/server.go
ethervoid: Metrics server is a different system that is decoupled from the server and I'm treating it like that. That's why it's outside the server package and acts aside receiving metrics. Imagine we have it like a microservice that receives and exposes metrics from different sources

metrics/transport.go
```diff
@@ -0,0 +1,47 @@
...
+// the requests and increase some metrics, cache
+// errors, etc based on the requests and responses
+type Transport struct {
+	Base    http.RoundTripper
+	metrics Provider
+}
+
+// NewTransport returns a transport using a provided http.RoundTripper as
+// the base and a metrics provider
+func NewTransport(base http.RoundTripper, metrics Provider) *Transport {
```
ethervoid, line 19: One thing. When we move the `Provider` to the server package this Provider dependency is going to need the server package. Is that right or makes sense to move the transport to the server package too (hopefully not).

// @isacikgoz @agnivade

agnivade: Yeah the problem is of package boundaries. If the Transport consumes the interface, then it cannot be in the source package which provides the implementation. And if transport uses the real implementation, then the client package needs to cast it back into the main implementation. So it's problematic both ways.

Yet another way might have been to include the transport and the roundtripper interface into the main interface itself. But I don't like that.

I think this is fine.

metrics/provider.go
```diff
@@ -0,0 +1,172 @@
...
+	cronTasksErrors   *prometheus.CounterVec
+
+	githubRequests    *prometheus.HistogramVec
+	githubCacheHits   *prometheus.CounterVec
+	githubCacheMisses *prometheus.CounterVec
+}
+
+// NewPrometheusProvider creates a new prometheus metrics provider
```
agnivade, line 38: Nit: period missing at the end of sentence.

metrics/provider.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+	provider.Registry.MustRegister(prometheus.NewProcessCollector(options))
+	provider.Registry.MustRegister(prometheus.NewGoCollector())
+
+	provider.httpRequestsDuration = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: metricsNamespace,
+			Subsystem: httpNamespace,
+			Name:      "requests",
+			Help:      "Received http requests.",
```
agnivade, line 55: Perhaps "HTTP requests by different categories." sounds better?

server/metrics.go
```diff
@@ -0,0 +1,74 @@
...
+
+import (
+	"net/http"
+	"strconv"
+	"time"
+)
+
+// MetricsProvider is the interface that exposes the communication with the metrics system
+// this interface should be implemented by the different providers we want to include
+type MetricsProvider interface {
```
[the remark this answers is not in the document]
ethervoid: I was waiting to know your oppinion about the cron task ones. If both of you see the value on them I'll add the code to update the metrics :angel:

Makefile
```diff
@@ -58,6 +58,7 @@ mocks:
 	mockgen -package mocks -destination store/mocks/pull_requests.go github.com/mattermost/mattermost-mattermod/store PullRequestStore
 	mockgen -package mocks -destination store/mocks/issue.go github.com/mattermost/mattermost-mattermod/store IssueStore
 	mockgen -package mocks -destination store/mocks/spinmint.go github.com/mattermost/mattermost-mattermod/store SpinmintStore
+	mockgen -package mocks -destination metrics/mocks/provider.go github.com/mattermost/mattermost-mattermod/metrics Provider
```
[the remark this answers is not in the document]
ethervoid: Yes, good point

metrics/metrics_test.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	prometheusModels "github.com/prometheus/client_model/go"
+	"github.com/stretchr/testify/require"
+)
+
+func TestMetrics(t *testing.T) {
+	provider := NewPrometheusProvider()
+	server := NewServer("12345", provider.Handler(), false)
+	server.Start()
+	time.Sleep(time.Second * 1)
```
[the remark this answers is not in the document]
ethervoid: To wait for the metrics server to start. Maybe I can reduce it to 500 ms o 200 ms

server/metrics.go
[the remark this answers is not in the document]
agnivade: I think they are valuable. More visibility never hurt anyone ;)

metrics/metrics_test.go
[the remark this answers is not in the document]
agnivade: I would just probably write a ping loop to check for availability and avoid hardcoded sleeps.

server/pull_request.go
```diff
@@ -427,12 +435,17 @@ func (s *Server) CheckPRActivity() {
...
 func (s *Server) CleanOutdatedPRs() {
 	mlog.Info("Cleaning outdated PRs in the mattermod database....")
-
+	start := time.Now()
 	ctx, cancel := context.WithTimeout(context.Background(), defaultCronTaskTimeout*time.Second)
-	defer cancel()
+	defer func() {
+		elapsed := float64(time.Since(start)) / float64(time.Second)
+		s.Metrics.ObserveCronTaskDuration("clean_outdated_prs", elapsed)
+		defer cancel()
```
agnivade, line 443: We can keep the `defer cancel()` outside as usual.

server/server_test.go
```diff
@@ -33,3 +36,53 @@ func TestPing(t *testing.T) {
 	_, err = time.ParseDuration(ping.Uptime)
 	require.NoError(t, err)
 }
+
+func TestGithubEvent(t *testing.T) {
```
agnivade, line 40: :100:

server/server.go
```diff
@@ -203,23 +222,33 @@ func (s *Server) ping(w http.ResponseWriter, r *http.Request) {
...
 	defer cancel()
+	defer func() {
+		elapsed := float64(time.Since(start)) / float64(time.Second)
+		statusCode := strconv.Itoa(w.(*responseWriterWrapper).StatusCode())
+		s.Metrics.ObserveHTTPRequestDuration(r.Method, "/pr_event", statusCode, elapsed)
+	}()
 	buf, err := ioutil.ReadAll(r.Body)
 	if err != nil {
 		mlog.Error("Failed to read body", mlog.Err(err))
+		w.WriteHeader(http.StatusBadRequest)
```
agnivade, line 237: :100:

ethervoid:
@agnivade  I’ve launch the tests 1k times and It didn’t fail so let’s try without it and if it fails then we add some watcher

agnivade:
Fine by me.

PR #175, opened 2020-07-17
[MM-26819] server: add some error propogation to handleIssueEvent

isacikgoz:
#### Summary
Added error propagation to the issue event handler. The ticket was originally for a logical error in the `server.githubEvent`, but it is taken care in #168 so I added a small refactor around error handling instead.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26819

### DIFF
--- a/server/issue.go
+++ b/server/issue.go
@@ -54,7 +54,7 @@ func (s *Server) checkIssueForChanges(ctx context.Context, issue *model.Issue) e
 		if !hadLabel {
 			mlog.Info("issue added label", mlog.Int("issue", issue.Number), mlog.String("label", label))
 			if err = s.handleIssueLabeled(ctx, issue, label); err != nil {
-			    return fmt.Errorf("could not handle issue label added: %w", err)
+				return fmt.Errorf("could not handle issue label added: %w", err)
 			}
 			hasChanges = true
 		}
--- a/server/issue.go
+++ b/server/issue.go
@@ -53,9 +53,8 @@ func (s *Server) checkIssueForChanges(ctx context.Context, issue *model.Issue) e
 
 		if !hadLabel {
 			mlog.Info("issue added label", mlog.Int("issue", issue.Number), mlog.String("label", label))
-			err = s.handleIssueLabeled(ctx, issue, label)
-			if err != nil {
-				return fmt.Errorf("could not handle issue label added: %w", err)
+			if err = s.handleIssueLabeled(ctx, issue, label); err != nil {
+			    return fmt.Errorf("could not handle issue label added: %w", err)
 			}
 			hasChanges = true
 		}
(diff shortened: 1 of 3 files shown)


server/issue.go
```diff
@@ -59,30 +53,31 @@ func (s *Server) checkIssueForChanges(ctx context.Context, issue *model.Issue) {
 
 		if !hadLabel {
 			mlog.Info("issue added label", mlog.Int("issue", issue.Number), mlog.String("label", label))
-			s.handleIssueLabeled(ctx, issue, label)
+			err = s.handleIssueLabeled(ctx, issue, label)
+			if err != nil {
+				return fmt.Errorf("could not handle issue label added: %w", err)
+			}
```
ethervoid, line 59: What about this?
```suggestion
			if err = s.handleIssueLabeled(ctx, issue, label); err != nil {
			    return fmt.Errorf("could not handle issue label added: %w", err)
			}
```

server/server.go
```diff
@@ -239,7 +241,10 @@ func (s *Server) githubEvent(w http.ResponseWriter, r *http.Request) {
 
 	eventData := EventDataFromJSON(ioutil.NopCloser(bytes.NewBuffer(buf)))
 	if eventData == nil || eventData.Action != "created" {
-		s.handleIssueEvent(ctx, event)
+		err = s.handleIssueEvent(ctx, event)
+		if err != nil {
+			mlog.Error(err.Error())
+		}
```
ethervoid, line 247:
```suggestion
		if err = s.handleIssueEvent(ctx, event); err != nil {
			mlog.Error(err.Error())
		}
```

server/issue.go
agnivade: We call this the Claudio coding style.

server/issue.go
```diff
@@ -5,47 +5,41 @@ package server
...
 
 	"github.com/mattermost/mattermost-mattermod/model"
 	"github.com/mattermost/mattermost-server/v5/mlog"
 )
 
-func (s *Server) handleIssueEvent(ctx context.Context, event *PullRequestEvent) {
+func (s *Server) handleIssueEvent(ctx context.Context, event *PullRequestEvent) error {
 	if event == nil || event.Issue == nil {
-		return
+		return errors.New("could not handle issue event: either event or issue is nil")
```
agnivade, line 18: I am wondering whether this should be considered as an error scenario?

isacikgoz: Actually I thought about this, from the function perspective it is an error because it requires these values to be set. Actually we should be sure about these values not to be nil before coming this point. So, the check was here and I deferred this to handler separation.

agnivade: I see. Makes sense.

PR #177, opened 2020-07-17
Containerize mattermod

khos2ow:
#### Summary

Containerizing `mattermod`. As a result two extra Makefile targets adde:

- `make docker`: to build the docker image
- `make push`: to push docker image to DockerHub

Example usage:

```bash
DOCKER_TAG=latest make docker push
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,29 @@
+FROM golang:1.14.2-alpine AS builder
+
+RUN apk add --update --no-cache ca-certificates bash make gcc musl-dev git openssh wget curl bzr
+
+WORKDIR /go/src/mattermod
+
+COPY go.mod .
+COPY go.sum .
+RUN go mod download
+
+COPY . .
+RUN CGO_ENABLED=0 make build
+
+################
+
+FROM alpine:3.12.0
+
+RUN apk --no-cache add ca-certificates
+
+COPY --from=builder /go/src/mattermod/dist/mattermod /usr/local/bin/
+
+WORKDIR /app
+
+RUN mkdir -p /app/logs && chown -R 1000:1000 /app/logs/
+
+USER 1000
+EXPOSE 8080
+
+ENTRYPOINT ["mattermod"]
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,11 @@ PACKAGES=$(shell go list ./...)
 .PHONY: all
 all: check-style test
 
+## Cleans workspace
+.PHONY: clean
+clean:
+	rm -rf dist/ out/
+
 ## Checks code style.
 .PHONY: check-style
 check-style: golangci-lint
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+COPY --from=builder /go/src/mattermod/dist/mattermod /usr/local/bin/
+
+WORKDIR /app
+
+RUN mkdir -p /app/logs && chown -R 1000:1000 /app/logs/
+
+USER 1000
+EXPOSE 8080
+
+ENTRYPOINT ["mattermod"]
```
ethervoid, line 29: Without configuration is going to fail even though I'm passing the config file:
```
$ docker run --name mattermod mattermost/mattermod:e5bbce0 -config config/config-mattermod.default.json
{"level":"error","msg":"unable to load server config","fields":[{"Key":"error","Type":25,"Integer":0,"String":"","Interface":{}}]}
```

Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM golang:1.14.2-alpine AS builder
+
+RUN apk add --update --no-cache ca-certificates bash make gcc musl-dev git openssh wget curl bzr
+
+WORKDIR /go/src/mattermod
+
+COPY go.mod .
```
ethervoid, line 7: Why don't do `COPY . .` and then `RUN go mod download`?

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+COPY . .
+RUN CGO_ENABLED=0 make build
+
+################
+
+FROM alpine:3.12.0
+
+RUN apk --no-cache add ca-certificates
+
+COPY --from=builder /go/src/mattermod/dist/mattermod /usr/local/bin/
```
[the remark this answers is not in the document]
khos2ow: Oh yes, this is the preferred approach (multi-stage docker build), basically the last stage is the actual image and layers, and you can do whatever needed in previous stages without worrying about bloating image with building/tooling which are not required in production.

Dockerfile
khos2ow: because frequency of changes in `go.mod` and `go.sum` are way lower than the rest of the codebase, so with this approach these three layers (`COPY go.mod`, `COPY go.sum` and `RUN go mod download`) do get cached properly and **only** will re-run again if there's actual changes in `go.mod` or `go.sum` or both.

Makefile
```diff
@@ -24,25 +29,32 @@ golangci-lint:
...
 	@echo test success
 
+## Builds mattermod.
+.PHONY: build
+build: clean
+	@echo Building
+	$(GO) build -o dist/mattermod ./cmd/mattermost-mattermod
+
+# Docker variables
+DEFAULT_TAG  ?= $(shell git describe --tags --exact-match 2>/dev/null || git rev-parse --short HEAD 2>/dev/null)
```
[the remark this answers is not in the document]
khos2ow: I originally put it up top, but then moved it down to be close to the targets use them. I don't have any preferences on ether though, only though if I left it up top might have given an impression this is needed or used globally in the Makefile but in reality it's only needed and used in two specific targets.

Dockerfile
khos2ow: The way you used it assumes the `config/config-mattermost.default.json` already exists in the container, which does not. The correct way to do it is to map a volume which contains the config.

```bash
docker run \
    -v /path/to/config-mattermod.default.json:/app/config/config.json \
    --name mattermod \
    mattermost/mattermod:e5bbce0 -config /app/config/config.json
````

Makefile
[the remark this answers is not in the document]
ethervoid: I don’t have a strong opinion either so we can leave it as it is

Dockerfile
ethervoid: yep given the complexity of the config file I think this is the way (insert Mandalorian gif here)

Dockerfile
ethervoid: Makes total sense

agnivade:
Is my review still needed? :)

PR #178, opened 2020-07-20
Add k8s manifests and dev tooling

khos2ow:
#### Summary

Add deployment manifests, developer local toolings guide.

### DIFF
--- a/deploy/base/db/deployment.yaml
+++ b/deploy/base/db/deployment.yaml
@@ -1,0 +1,92 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: mysql
+spec:
+  strategy:
+    type: Recreate
+  selector:
+    matchLabels:
+      app: mysql
+  template:
+    metadata:
+      labels:
+        app: mysql
+    spec:
+      serviceAccountName: mysql
+      securityContext:
+        fsGroup: 999
+        runAsUser: 999
+      restartPolicy: Always
+      initContainers:
+        - name: "remove-lost-found"
+          image: "busybox:1.32"
+          imagePullPolicy: "IfNotPresent"
+          resources:
+            requests:
+              cpu: 10m
+              memory: 10Mi
+          command: ["rm", "-fr", "/var/lib/mysql/lost+found"]
+          volumeMounts:
+            - name: data
+              mountPath: /var/lib/mysql
+      containers:
+        - name: mysql
+          image: "mysql:5.7.30"
+          imagePullPolicy: "IfNotPresent"
+          resources:
+            limits:
+              cpu: 500m
+              memory: 512Mi
+            requests:
+              cpu: 100m
+              memory: 256Mi
+          env:
+            - name: MYSQL_ROOT_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: mysql
+                  key: mysql-root-password
+            - name: MYSQL_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: mysql
+                  key: mysql-password
+                  optional: true
+            - name: MYSQL_USER
+              value: "mattermod"
+            - name: MYSQL_DATABASE
+              value: "mattermod"
+          ports:
+            - name: mysql
+              containerPort: 3306
+          livenessProbe:
+            exec:
+              command:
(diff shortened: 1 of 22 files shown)


khos2ow:
@agnivade @ethervoid FYI, this will be the new way we're going to deploy mattermod, but on the other hand you can also use the same local workflow (described in README) for local development and testing on this too.

server/server.go
```diff
@@ -215,7 +217,17 @@ func (s *Server) Tick() {
...
-	err := json.NewEncoder(w).Encode(pingResponse{Uptime: uptime})
+	err := json.NewEncoder(w).Encode(pingResponse{Uptime: uptime, Healthy: "OK!"})
+	if err != nil {
+		mlog.Error("Failed to write ping", mlog.Err(err))
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+}
+
+func (s *Server) healthz(w http.ResponseWriter, r *http.Request) {
```
agnivade, line 228: Wondering if there a reason we are duplicating the code?

agnivade:
@ashishbhate @isacikgoz :point_up_2: in case any of you want to use k8s locally.

server/server.go
metanerd: Do you mean to use the same controller function, no? 
```
"/healthz", s.ping
```

agnivade: Yeah, same controller function, or same endpoint. Either works.

khos2ow: I started creating `/healthz` endpoint and controller function and then noticed that `/` actually does nothing and only shows uptime. I didn't want to use `/` as endpoint or its underlying controller function for health checking (and liveness and readiness probes) because 1) `/healthz` is a very known endpoint in this space 2) at later time `/` might get changed.

Although I should have created `healthResponse` struct and not reuse `pingResponse` (which I'm gonna change).

agnivade: IMO, _if_ we change the `/` endpoint, then we can easily create a separate handler at that point. Pre-emptively duplicating code thinking of a future change which hasn't happened yet seems redundant.

khos2ow: Fair enough, I'll point `/healthz` endpoint to existing ping controller.

deploy/base/db/deployment.yaml
```diff
@@ -0,0 +1,92 @@
...
+            requests:
+              cpu: 10m
+              memory: 10Mi
+          command: ["rm", "-fr", "/var/lib/mysql/lost+found"]
+          volumeMounts:
+            - name: data
+              mountPath: /var/lib/mysql
+      containers:
+        - name: mysql
+          image: "mysql:5.7.30"
```
metanerd, line 36: Wow, that's specific! Why not mysql:5.7 for getting patches?

deploy/base/db/secrets.yaml
```diff
@@ -0,0 +1,9 @@
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: mysql
+type: Opaque
+data:
+  mysql-root-password: "cGFzc3dvcmQ="
```
metanerd, line 8: Can we use other more standard garbled ones?

khos2ow: I wasn't aware of such other standard. Examples?

deploy/base/db/deployment.yaml
khos2ow: I like to pinpoint on very specific version, to have absolute visibility and control of when and how to upgrade, otherwise a pod might die and come back up with an upgraded version and all of a sudden something might have gotten changed.

deploy/base/db/secrets.yaml
metanerd: Nvm, forgot it was 64encoded :facepalm:

deploy/base/db/deployment.yaml
ethervoid: Sadly MySQL sometimes include breaking changes in the patch versions so it's better to be sure

khos2ow: That's true, but bear in mind we've changed the logic a bit. It means we're using mysql which we deploy ourselves in dev mode and use managed database provided by cloud provider for production.

PR #179, opened 2020-07-21
MM-26927: Use a recovery middleware

agnivade:
#### Ticket link

https://mattermost.atlassian.net/browse/MM-26927

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -157,7 +157,7 @@ func (s *Server) withRecovery(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		defer func() {
 			if x := recover(); x != nil {
-				mlog.Error("panic occurred", mlog.String("url", r.URL.String()), mlog.Any("error", x))
+				mlog.Error("recovered from a panic", mlog.String("url", r.URL.String()), mlog.Any("error", x))
 			}
 		}()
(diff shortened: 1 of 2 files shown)


agnivade:
Why yes of course :) Thanks for the push.

server/server.go
```diff
@@ -152,6 +153,17 @@ func (s *Server) RefreshMembers() {
 	s.OrgMembers = members
 }
 
+func (s *Server) withRecovery(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		defer func() {
+			if x := recover(); x != nil {
+				mlog.Error("recovered from a panic", mlog.String("url", r.URL.String()), mlog.Any("error", x))
```
isacikgoz, line 160: Shall we also print the request body and header? It would be great to see which request caused the panic :)

agnivade: I think that would pollute the logs too much. And plus you need another NopCloser to read the body again.

Header is unlikely to contain any helpful information. The thing to check would be the body. I am guessing that the panic trace should be enough for us to figure out the bug.

isacikgoz: I see, let's continue with this then. 👍

agnivade:
@metanerd - Please deploy at your leisure.

metanerd:
@agnivade done

PR #184, opened 2020-07-22
[MM-27175] server: add middleware for validation

isacikgoz:
#### Summary
Adds a middleware for validating incoming requests.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27175


server/server.go
```diff
@@ -164,6 +165,44 @@ func (s *Server) withRecovery(next http.Handler) http.Handler {
...
 
+func (s *Server) withValidation(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		start := time.Now()
+		w = newWrappedWriter(w)
+
+		defer func() {
+			elapsed := float64(time.Since(start)) / float64(time.Second)
+			statusCode := strconv.Itoa(w.(*responseWriterWrapper).StatusCode())
+			s.Metrics.ObserveHTTPRequestDuration(r.Method, "/pr_event", statusCode, elapsed)
```
ashishbhate, line 176: How about using `r.URL.Path` (not sure if its the best way to get the URL basepath) instead of hardcoding `/pr_event`?

This could end up being missed when we split up the routes for different webhook events.

isacikgoz: Makes sense 👍

server/server.go
```diff
@@ -164,6 +165,44 @@ func (s *Server) withRecovery(next http.Handler) http.Handler {
 	})
 }
 
+func (s *Server) withValidation(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		start := time.Now()
+		w = newWrappedWriter(w)
+
+		defer func() {
+			elapsed := float64(time.Since(start)) / float64(time.Second)
```
agnivade, line 174: Oh I actually misunderstood you. I thought you meant that _all_ metrics code will be moved to middleware. That's why I said it can be called from a lot of places. Now I realize you were just talking about request duration. In that case, I think a separate middleware is fine. Even the name `WithValidation` is misleading because it does more than one thing. Apologies for the double work.

Let's also move the `withValidation` method inside validate_signature.go and un-export `ValidateSignature`.

And this is a great method to write some tests since this has become more modular. Let's add them too.

PR #186, opened 2020-07-22
[MM-27017] server: add new endpoint for issues webhook

isacikgoz:
#### Summary
PR adds `issueEventHandler` for issues webhook. Should be merged after #184.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27017


server/issue_handler.go
```diff
@@ -0,0 +1,108 @@
...
+)
+
+func (s *Server) issueEventHandler(w http.ResponseWriter, r *http.Request) {
+	ctx, cancel := context.WithTimeout(context.Background(), defaultRequestTimeout*time.Second)
+	defer cancel()
+
+	event, err := PullRequestEventFromJSON(r.Body)
+	if err != nil {
+		mlog.Error("could not parse issue event", mlog.Err(err))
+		w.WriteHeader(http.StatusBadRequest)
```
agnivade, line 24: This is just a general thing and not related to this PR, but I note that in case of error scenarios, we are just writing the status code and returning. No body is written. Wondering if it makes sense to use the `http.Error` function to write out the error in the response along with the status code? That will allow us to view the error response in the GitHub console too.

server/issue.go
```diff
@@ -9,7 +9,6 @@ import (
 	"fmt"
```
agnivade, line 9: This entire file will go away once we migrate the issues handler right?

server/issue_handler.go
```diff
@@ -0,0 +1,108 @@
...
+
+	"github.com/mattermost/mattermost-mattermod/model"
+	"github.com/mattermost/mattermost-server/v5/mlog"
+)
+
+func (s *Server) issueEventHandler(w http.ResponseWriter, r *http.Request) {
+	ctx, cancel := context.WithTimeout(context.Background(), defaultRequestTimeout*time.Second)
+	defer cancel()
+
+	event, err := PullRequestEventFromJSON(r.Body)
```
agnivade, line 21: I am a bit confused here. I am assuming we are migrating the issues webhook here. Shouldn't we use the `EventDataFromJSON` function here? Even that does not match the exact format [here]( https://docs.github.com/en/developers/webhooks-and-events/webhook-events-and-payloads#issues).

I would just have 3 separate `xxFromJSON` functions for 3 separate events. It's much clearer this way.
- IssueFromJSON
- IssueCommentFromJSON
- PullRequestFromJSON

server/issue_handler.go
```diff
@@ -0,0 +1,108 @@
...
+	if err != nil {
+		mlog.Error("could not parse issue event", mlog.Err(err))
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+
+	parts := strings.Split(event.Issue.GetHTMLURL(), "/")
```
agnivade, line 28: Should we use getters everywhere? `event.GetIssue().GetHTMLURL()` ?

server/issue_handler.go
```diff
@@ -0,0 +1,108 @@
...
+	if len(parts) < 4 {
+		mlog.Error("incorrect pattern for issue url")
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+
+	mlog.Info("handle issue event", mlog.String("repoUrl", *event.Issue.HTMLURL), mlog.String("Action", event.Action), mlog.Int("PRNumber", event.PRNumber))
```
agnivade, line 35: Getters?

server/issue_handler.go
```diff
@@ -0,0 +1,108 @@
...
+	issue, err := s.GetIssueFromGithub(ctx, parts[len(parts)-4], parts[len(parts)-3], event.Issue)
```
agnivade, line 36: Can we add a comment to clarify what does `parts[len(parts)-4]` and `parts[len(parts)-3]` signify?

server/issue_handler.go
```diff
@@ -0,0 +1,108 @@
...
+	event, err := PullRequestEventFromJSON(r.Body)
+	if err != nil {
+		mlog.Error("could not parse issue event", mlog.Err(err))
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+
+	parts := strings.Split(event.Issue.GetHTMLURL(), "/")
+	if len(parts) < 4 {
+		mlog.Error("incorrect pattern for issue url")
```
agnivade, line 30: Let's log the URL?

server/issue_handler.go
```diff
@@ -0,0 +1,108 @@
...
+
+func (s *Server) checkIssueForChanges(ctx context.Context, issue *model.Issue) error {
+	oldIssue, err := s.Store.Issue().Get(issue.RepoOwner, issue.RepoName, issue.Number)
+	if err != nil {
+		return err
+	}
+
+	if oldIssue == nil {
+		_, err = s.Store.Issue().Save(issue)
+		return err
```
agnivade, line 57: bug :)

server/issue_handler.go
```diff
@@ -0,0 +1,108 @@
...
+				return fmt.Errorf("could not handle issue label added: %w", err)
+			}
+			hasChanges = true
+		}
+	}
+
+	if hasChanges {
+		mlog.Info("issue has changes", mlog.Int("issue", issue.Number))
+		_, err = s.Store.Issue().Save(issue)
+		return err
```
agnivade, line 84: Not a bug technically, but fragile code and will break the moment we add more code after this. Let's fix this.

server/issue_handler_test.go
```diff
@@ -0,0 +1,167 @@
...
+		Labels:    []string{},
+		Number:    1,
+	}
+
+	ts := httptest.NewServer(http.HandlerFunc(s.issueEventHandler))
+	defer ts.Close()
+
+	ctxInterface := reflect.TypeOf((*context.Context)(nil)).Elem()
+
+	t.Run("Should fail with no body", func(t *testing.T) {
```
agnivade, line 74: Can we also add a test where the CreateComment method is called.

server/server.go
```diff
@@ -91,6 +91,7 @@ func New(config *Config, metrics MetricsProvider) (*Server, error) {
 
 	r := mux.NewRouter()
 	r.HandleFunc("/", s.ping).Methods(http.MethodGet)
+	r.HandleFunc("/issue", s.issueEventHandler).Methods(http.MethodPost)
```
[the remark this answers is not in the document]
metanerd: I would go with issue. Because it is an event and it gets triggered when an issue is in some way touched.

server/issue.go
isacikgoz: Yes, my plan to consolidate handler related code to `xxx_handler.go` in the long run.

server/issue_handler.go
isacikgoz: You're right, this is confusing. Actually I wanted to migrate gradually to not break things since we are shipping the code immediately :) However, we can have a new `type` for this handler which only includes `action` and `issue` itself. In fact, we can handle this in parsing stage and use our `model.Issue` for the rest of the code. Will update this accordingly.

server/issue_handler.go
isacikgoz: How? 😅

server/server.go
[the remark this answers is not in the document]
agnivade: Sounds good.

server/issue_handler.go
agnivade: Lol, we are just directly returning from here unconditionally :)

server/github.go
```diff
@@ -87,13 +87,13 @@ func (s *Server) GetPullRequestFromGithub(ctx context.Context, pullRequest *gith
...
-		RepoOwner: repoOwner,
-		RepoName:  repoName,
-		Number:    *ghIssue.Number,
-		Username:  *ghIssue.User.Login,
-		State:     *ghIssue.State,
+		RepoOwner: ghIssue.GetRepository().GetOwner().GetLogin(),
+		RepoName:  ghIssue.GetRepository().GetName(),
+		Number:    ghIssue.GetNumber(),
+		Username:  ghIssue.GetUser().GetLogin(),
+		State:     ghIssue.GetState(),
```
isacikgoz, line 96: This change should be carefully reviewed, I expect these were redundant. That's why I removed them. However, it's not so clear in the [docs](https://docs.github.com/en/developers/webhooks-and-events/github-event-types).

PR #192, opened 2020-07-23
Skip webhook SHA validation for /healthz

khos2ow:
#### Summary

#184 added middleware validation which this PR skips that only for `/healthz` endpoint.

### DIFF
--- a/server/validate_signature.go
+++ b/server/validate_signature.go
@@ -16,6 +16,9 @@ import (
 
 func (s *Server) withValidation(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.RequestURI == "/healthz" {
+			next.ServeHTTP(w, r)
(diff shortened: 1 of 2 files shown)


agnivade:
@khos2ow - Please wait for 2 reviewers to give their approvals before merging the PR. If something is breaking production, we should revert code rather than quickly pushing more code. It just leads to more instability and hacks like this IMO. It is fine for this case, but ideally we should have done what @isacikgoz has suggested.

server/validate_signature.go
```diff
@@ -16,6 +16,9 @@ import (
 
 func (s *Server) withValidation(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.RequestURI == "/healthz" {
+			next.ServeHTTP(w, r)
```
agnivade, line 20: We should return after this line.

PR #195, opened 2020-07-24
Add cherry_pick.sh, git and hub to the image

khos2ow:
#### Summary

Add missing `cherry_pick.sh` to the image at `/app/scripts/cherry_pick.sh`

### DIFF
--- a/server/cherry_pick.go
+++ b/server/cherry_pick.go
@@ -108,7 +108,7 @@ func (s *Server) doCherryPick(ctx context.Context, version string, milestoneNumb
 	}
 	releaseBranch := fmt.Sprintf("upstream/%s", version)
 	repoFolder := fmt.Sprintf("/home/ubuntu/git/mattermost/%s", pr.RepoName)
-	cmd := exec.Command("/home/ubuntu/git/devops/cherry-pick.sh", releaseBranch, strconv.Itoa(pr.Number), pr.MergeCommitSHA)
+	cmd := exec.Command("/app/scripts/cherry-pick.sh", releaseBranch, strconv.Itoa(pr.Number), pr.MergeCommitSHA)
 	cmd.Dir = repoFolder
 	cmd.Env = append(
 		os.Environ(),
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,6 +18,7 @@ FROM alpine:3.12.0
 RUN apk --no-cache add ca-certificates
 
 COPY --from=builder /go/src/mattermod/dist/mattermod /usr/local/bin/
+COPY --from=builder /go/src/mattermod/hack/cherry-pick.sh /app/scripts/cherry-pick.sh
 
 WORKDIR /app
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,6 +2,12 @@ FROM golang:1.14.2-alpine AS builder
 
 RUN apk add --update --no-cache ca-certificates bash make gcc musl-dev git openssh wget curl bzr
 
+ENV HUB_VERSION 2.14.2
+WORKDIR /opt/hub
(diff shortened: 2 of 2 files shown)


server/cherry_pick.go
```diff
@@ -108,7 +108,7 @@ func (s *Server) doCherryPick(ctx context.Context, version string, milestoneNumb
 	}
 	releaseBranch := fmt.Sprintf("upstream/%s", version)
 	repoFolder := fmt.Sprintf("/home/ubuntu/git/mattermost/%s", pr.RepoName)
-	cmd := exec.Command("/home/ubuntu/git/devops/cherry-pick.sh", releaseBranch, strconv.Itoa(pr.Number), pr.MergeCommitSHA)
+	cmd := exec.Command("/app/scripts/cherry-pick.sh", releaseBranch, strconv.Itoa(pr.Number), pr.MergeCommitSHA)
```
agnivade, line 111: We seriously need to move from this hardcoded path to something that is coming from the config. @ashishbhate - makes sense to track that in a new ticket?

khos2ow: Anything config, I completely agree, but this is the underlying foundation of the image and I don't think this is needed to get parameterized or put in config.json.

khos2ow:
@ashishbhate so mattermod does actually git clone repos locally?

server/cherry_pick.go
ashishbhate: Yep, I made a ticket yesterday when trying to get mattermod working on my local machine. https://mattermost.atlassian.net/browse/MM-27192

ashishbhate:
>@ashishbhate so mattermod does actually git clone repos locally?

mattermod does not clone the repos. It expects the various repos to already exist on the disk (https://github.com/mattermost/mattermost-mattermod/blob/master/server/cherry_pick.go#L110)

mattermod does git operations on them through the `git` command and `hub` command.

ashishbhate:
Just to be clear, only the cherry-pick part of mattermod needs the local repo checkouts and `git` and `hub`. The rest of the functions (like merge) use the GitHub API to do it directly in Github. 

The Github API does not have cherry-picking. There are ways to emulate cherry-picking using the Github API, by updating refs etc., but these are not reliable and often confusing.

khos2ow:
1) what happens when mattermod finishes with a cherry-pick of a repo? I saw it goes back to `master`, does that mean it leaves it there on disk?
2) what happens if a repo doesn't exist on the file system?

ashishbhate:
> 1. what happens when mattermod finishes with a cherry-pick of a repo? I saw it goes back to master, does that mean it leaves it there on disk?

Yes, the repo needs to stay on the disk.

>what happens if a repo doesn't exist on the file system?

The code complains that it cannot cd into that folder. https://github.com/mattermost/mattermost-mattermod/blob/master/server/cherry_pick.go#L110-L118


>does mattermod handle concurrency on operations on single repo? I mean if two cherry-pick requests come at the same time, are they queued to be operated on the same folder in sequence?

Nope, the second one fails. The bash script checks if git operations are in progress. If there is already a cherry-pick in progress it errors out.

khos2ow:
Ok, then I'm gonna attach a data drive to the container (at `/app/data` or `/data`) and then we can have the cherry_pick.go points to `/app/data/mattermost/` (I don't think `/git/` is really needed).

> The code complains that it cannot cd into that folder

So what's the workaround? manually cloning the repo on disk?

ashishbhate:
>So what's the workaround? manually cloning the repo on disk?

yep, for now manually cloning repo on disk. 
I'll make a ticket for an initialization step when mattermod starts, to have mattermod clone the repos to disk. That way devops wont have to do anything special (especially if we add more repos for mattermod to monitor).

We dont depend on any particular version of git or hub (https://github.com/github/hub/releases), so the latest versions for both work.

khos2ow:
@agnivade @ashishbhate this is fine now, I'm gonna add data disk to the image in a separate PR.

PR #196, opened 2020-07-24
Mount /app/repos persistent volume into container

khos2ow:
#### Summary

Add a persistent volume to deployment for local clones of repositories on disk.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,6 +32,7 @@ COPY --from=builder /go/src/mattermod/hack/cherry-pick.sh /app/scripts/cherry-pi
 
 WORKDIR /app
 
+RUN mkdir -p /app/repos && chown -R 1000:1000 /app/repos/
 RUN mkdir -p /app/logs && chown -R 1000:1000 /app/logs/
 
 USER 1000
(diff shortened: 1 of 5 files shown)


agnivade:
I don't have any knowledge on how any of this works :P Perhaps @metanerd is a better person to review this?

ashishbhate:
>I don't have any knowledge on how any of this works :P Perhaps @metanerd is a better person to review this?

ditto

khos2ow:
Basically the most important part is:

```diff
-	repoFolder := fmt.Sprintf("/home/ubuntu/git/mattermost/%s", pr.RepoName)
+	repoFolder := fmt.Sprintf("/app/repos/%s", pr.RepoName)
```

which is a hardcoded path to a data disk (which happens to be mounted into the container) to store the repository clones.

khos2ow:
Originally I wanted to use `ReadWriteMany` access mode but there's a limitation of using that on AWS, for time being I've converted to `ReadWriteOnce` (which is sufficient enough) to unblock prod and later I'm gonna figure out a way to fall back on _Many_.

PR #197, opened 2020-07-27
[MM-27018] server/pr_comment_handler: add new pr_comment webhook handler

isacikgoz:
#### Summary
PR adds `prCommentEventHandler` for pr comments webhook. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27018


server/pr_comment_handler.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+func (s *Server) prCommentEventHandler(w http.ResponseWriter, r *http.Request) {
+	ctx, cancel := context.WithTimeout(context.Background(), defaultRequestTimeout*time.Second)
+	defer cancel()
+
+	ev, err := prCommentEventFromJSON(r.Body)
+	if err != nil {
+		mlog.Error("could not parse pr comment event", mlog.Err(err))
+		http.Error(w, err.Error(), http.StatusBadRequest)
```
ethervoid, line 33: Noice, didn't know about it :+1:

server/pr_comment_handler.go
```diff
@@ -0,0 +1,124 @@
...
+
+	pr, err := s.getPRFromEvent(ctx, ev)
+	if err != nil {
+		mlog.Error("Error getting PR from Comment", mlog.Err(err))
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+	var commenter string
+	if ev.Comment != nil {
+		commenter = ev.Comment.GetUser().GetLogin()
```
ethervoid, line 46: GetUser() could return nil as well so we should include that check in the conditional

server/pr_comment_handler_test.go
```diff
@@ -0,0 +1,133 @@
...
+				Login: &login,
+			},
+		},
+	}
+
+	ts := httptest.NewServer(http.HandlerFunc(s.prCommentEventHandler))
+	defer ts.Close()
+
+	ctxInterface := reflect.TypeOf((*context.Context)(nil)).Elem()
+
```
ethervoid, line 84: We should include tests when there isn't issue, comment or repository

server/pr_comment_handler.go
isacikgoz: It's @agnivade to push this 🚀

server/github.go
```diff
@@ -38,9 +38,7 @@ func (s *Server) GetPullRequestFromGithub(ctx context.Context, pullRequest *gith
 		MaintainerCanModify: sql.NullBool{Bool: pullRequest.GetMaintainerCanModify(), Valid: true},
 	}
 
-	if pullRequest.Head.Repo != nil {
```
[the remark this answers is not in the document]
isacikgoz: If it's nil, it will still return empty string and the value won't change because it's still initialized empty.

server/server.go
```diff
@@ -94,6 +94,7 @@ func New(config *Config, metrics MetricsProvider) (*Server, error) {
 
 	webhooks := r.PathPrefix("/webhooks").Subrouter()
 	webhooks.HandleFunc("/issue", s.issueEventHandler).Methods(http.MethodPost)
+	webhooks.HandleFunc("/pr_comment", s.prCommentEventHandler).Methods(http.MethodPost)
```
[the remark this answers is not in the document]
isacikgoz: I don't think there will be much work than trying to maintain this. And we can have much better separation of the problems.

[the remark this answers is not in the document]
ethervoid: Take into account that there are a good amount of events but if you have discussed this and is ok I'm not going to interfere

server/github.go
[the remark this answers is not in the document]
ethervoid: Nothing more to say https://play.golang.org/p/1kuQe3F8I_V

server/pr_comment_handler.go
ethervoid: Solved above

server/server.go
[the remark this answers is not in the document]
agnivade: Actually, I just came to know about the `X-GitHub-Event` today :facepalm:

Regarding number of events, there are just 3 of them. So it's not that big a deal to maintain from the code.

But the changing of webhooks for every repository is indeed a concern, and I was thinking the same. But at that time, I didn't know about the header :(

Let me ask this in the channel.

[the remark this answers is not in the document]
agnivade: Actually, I have another idea. We can use the GH API to create and update the webhooks in all the repos. That is a better and automated solution. Will discuss this in the channel.

server/pr_comment_handler_test.go
```diff
@@ -0,0 +1,133 @@
...
+	t.Run("Should fail with no body", func(t *testing.T) {
```
agnivade, line 85: I would also like a test where we differentiate whether the comment is coming from a PR or an issue and skip calling `getPRFromEvent` if that's the case.

server/issue_comment_handler.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	pr, err := s.getPRFromIssueCommentEvent(ctx, ev)
+	if err != nil {
+		mlog.Error("Error getting PR from Comment", mlog.Err(err))
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+	var commenter string
+	if ev.Comment != nil {
```
agnivade, line 50: Is this needed? We are already validation that the comment is not-nil in `issueCommentEventFromJSON`.

server/issue_comment_handler_test.go
```diff
@@ -0,0 +1,165 @@
...
+	number := 1
+	state := "a-state"
+	login := "mattertest"
+	name := "mattermod"
+	body := "some-text"
+
+	event := issueCommentEvent{
+		Repository: &github.Repository{
+			Owner: &github.User{
+				Login: &login,
```
agnivade, line 58: Nit: Instead of the `&var` hack, we can use `github.NewString`.

PR #200, opened 2020-07-28
MM-27329: Fix incorrect error check

agnivade:
#### Ticket link

https://mattermost.atlassian.net/browse/MM-27329


agnivade:
I have verified this works in my local setup. In fact, running the local setup is how I found this bug :P

PR #202, opened 2020-07-28
Add proper git configuration for runtime

khos2ow:
#### Summary

Add required git configuration for mattermod runtime operation, namely cherry-pick process.


agnivade:
Is there a way to test this without deploying to prod? :P

khos2ow:
With the current setup, no. But you can locally make the docker and `run -ti` to make sure the `/app/.ssh` folder is created.

On the other hand the same (but limited) type of config can be added to `overlays/dev` too, but I'm not sure how useful that would be.

PR #203, opened 2020-07-28
server: fix a regression introduced in #186

isacikgoz:
#### Summary
Fixes a regression caused in #186. The `event.Issue == nil` check was avoiding pr to stop falling through.


server/issue_handler.go
```diff
@@ -33,6 +33,7 @@ func (s *Server) issueEventHandler(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
 	}
+	defer r.Body.Close()
```
agnivade, line 36: This seems unrelated?

isacikgoz: Not specifically related with the check but it's an error haven't been discovered yet :)

agnivade: I don't believe we need to close the body in server response handlers. This is only applicable for clients making the requests.

> // The Server will close the request body. The ServeHTTP
    // Handler does not need to.

server/github_pr.go
```diff
@@ -61,18 +60,14 @@ func (d *EventData) HasUpdateBranch() bool {
...
 
-func PullRequestEventFromJSON(data io.Reader) (*PullRequestEvent, error) {
+func PullRequestEventFromJSON(data io.Reader) *PullRequestEvent {
 	decoder := json.NewDecoder(data)
 	var event PullRequestEvent
 	if err := decoder.Decode(&event); err != nil {
-		return nil, err
-	}
-
-	if event.Issue == nil {
```
khos2ow, line 71: I don't see why removing this would have any effect. If `event.Isssue` is nil then the `err` would be non-nil so `if err == nil && event.PRNumber != 0 {` block would never get executed.

server/issue_handler.go
isacikgoz: I see, didn't know that. Thanks 👍

server/github_pr.go
agnivade: Not sure I understand your point completely. If event.Issue is nil, we were returning an error previously, which prevented the `if err == nil && event.PRNumber != 0 {` block from getting executed. Now it will.

Perhaps I am missing something?

khos2ow: I have the impression if `event.Issue` is not nil then `event.PRNumber` is `0`. Isn't this the case?

agnivade: Not really. It's not mutually exclusive. The `Issue` field might be there for both scenarios. But the `PRNumber` will only be there if it's a PR.

It's super confusing, and that's why we are moving to separate endpoints.

khos2ow: Oh I see, it makes sense then!

PR #205, opened 2020-07-29
server: fix gh issue service error if there isnt repo field in issue

isacikgoz:
#### Summary
Yet another regression found during local testing for #186. In some cases GitHub does not adding Repository field for issues in webhook requests. In those cases, we try to parse the `htmlURL` to get repo name and repo owner.

### DIFF
--- a/server/issue_handler_test.go
+++ b/server/issue_handler_test.go
@@ -96,6 +96,24 @@ func TestIssueEventHandler(t *testing.T) {
 		require.Equal(t, http.StatusBadRequest, resp.StatusCode)
 	})
 
+	t.Run("Should fail incorrect url", func(t *testing.T) {
+		url := "http://someaddr.com"
+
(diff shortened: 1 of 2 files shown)


server/issue_handler_test.go
```diff
@@ -96,6 +96,24 @@ func TestIssueEventHandler(t *testing.T) {
 		require.Equal(t, http.StatusBadRequest, resp.StatusCode)
 	})
 
+	t.Run("Should fail incorrect url", func(t *testing.T) {
```
agnivade, line 99: Can we also add a test where repo owner and name is nil but HTMLUrl is set, and verify that `ListLabelsByIssue` is being called with the correct values?

agnivade:
Will do some testing with the other PRs.

agnivade:
Tested and looks good!

PR #206, opened 2020-07-29
server/circleci: make triggerCircleCiIfNeeded method return error

isacikgoz:
#### Summary
We are already logging in the beginning of the method. And we will log if there is any error during the call. Also, we should add some tests after [MM-26704](https://mattermost.atlassian.net/browse/MM-26704) merged.


server/circleci.go
```diff
@@ -20,42 +20,38 @@ import (
...
-		return
+		return nil
 	}
 
 	// Checking if the repo have circleci setup
 	builds, err := s.CircleCiClient.ListRecentBuildsForProjectWithContext(ctx, circleci.VcsTypeGithub, pr.RepoOwner, pr.RepoName, "master", "", 5, 0)
 	if err != nil {
-		mlog.Error("listing the circleci project", mlog.String("repo", pr.RepoName), mlog.Int("pr", pr.Number), mlog.String("Fullname", pr.FullName), mlog.Err(err))
-		return
+		return fmt.Errorf("could not list the circleci builds for project: %w", err)
```
ethervoid, line 35: Do we need to wrap the error? AFAIK we need to wrap errors if we want to check with `errors.Is` or something like that but maybe I'm wrong

server/circleci.go
```diff
@@ -67,10 +63,10 @@ func (s *Server) triggerCircleCiIfNeeded(ctx context.Context, pr *model.PullRequ
 
 	err = s.CircleCiClient.BuildByProjectWithContext(ctx, circleci.VcsTypeGithub, pr.RepoOwner, pr.RepoName, opts)
 	if err != nil {
-		mlog.Error("Error triggering circleci", mlog.String("repo", pr.RepoName), mlog.Int("pr", pr.Number), mlog.String("Fullname", pr.FullName), mlog.Err(err))
-		return
+		return fmt.Errorf("could not trigger circleci: %w", err)
 	}
-	mlog.Info("Triggered circleci", mlog.String("repo", pr.RepoName), mlog.Int("pr", pr.Number), mlog.String("fullname", pr.FullName))
```
agnivade, line 73: I'd prefer to keep this at the call site at a debug level. It's a useful debug info,

server/circleci.go
```diff
@@ -20,42 +20,38 @@ import (
 	"github.com/mattermost/go-circleci"
 )
 
-func (s *Server) triggerCircleCiIfNeeded(ctx context.Context, pr *model.PullRequest) {
+func (s *Server) triggerCircleCiIfNeeded(ctx context.Context, pr *model.PullRequest) error {
```
agnivade, line 23: Let's rename to "..CircleCI.." while we're here.

server/circleci.go
agnivade, line 35: CircleCI.

server/pull_request.go
```diff
@@ -30,7 +30,11 @@ func (s *Server) handlePullRequestEvent(ctx context.Context, event *PullRequestE
 		if err := s.handleCheckCLA(ctx, pr); err != nil {
 			mlog.Error("Unable to check CLA", mlog.Err(err))
 		}
-		s.triggerCircleCiIfNeeded(ctx, pr)
+
+		if err := s.triggerCircleCiIfNeeded(ctx, pr); err != nil {
```
agnivade, line 34: Let's also dump the repoName, prNumber and fullName in the error fields.

server/circleci.go
isacikgoz: I believe it's a good practice to pass errors upwards. Even if we don't unwrap it or check for specific error at caller site, we are chaining the errors and get a nice trace.

server/pull_request.go
```diff
@@ -30,7 +30,12 @@ func (s *Server) handlePullRequestEvent(ctx context.Context, event *PullRequestE
 		if err := s.handleCheckCLA(ctx, pr); err != nil {
 			mlog.Error("Unable to check CLA", mlog.Err(err))
 		}
-		s.triggerCircleCiIfNeeded(ctx, pr)
+
+		if err := s.triggerCircleCIIfNeeded(ctx, pr); err != nil {
+			mlog.Error("Unable to trigger CircleCI", mlog.String("repo", pr.RepoName), mlog.Int("pr", pr.Number), mlog.String("fullname", pr.FullName), mlog.Err(err))
+		}
+		mlog.Debug("Triggered CircleCI", mlog.String("repo", pr.RepoName), mlog.Int("pr", pr.Number), mlog.String("fullname", pr.FullName))
```
agnivade, line 37: This needs to be in an else condition. Same applies to the 2 instances below.

server/circleci.go
ethervoid: Depends. I read about it [here](https://blog.golang.org/go1.13-errors) in the "Whether to wrap" section. I thought that is a good practice to do it always but is not that way as it looks. Just wanted to give a heads up about it :D

server/pull_request.go
isacikgoz: 🤦

PR #208, opened 2020-07-29
Expose /metrics endpoint to be scraped later on

khos2ow:
#### Summary

Exposing `/metrics` endpoint internally for Promethues to scrape them later on.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,6 +50,6 @@ RUN for d in .ssh repos logs; do \
     done
 
 USER mattermod
-EXPOSE 8080
+EXPOSE 8080 9000
 
 ENTRYPOINT ["mattermod"]
--- a/Tiltfile
+++ b/Tiltfile
@@ -2,4 +2,4 @@ docker_build('mattermost/mattermod', '.', dockerfile='Dockerfile')
 
 k8s_yaml(kustomize('./deploy/overlays/dev'))
 
-k8s_resource('mattermod', port_forwards=8080)
+k8s_resource('mattermod', port_forwards=[8080, 9000])
--- a/config/config-mattermod.default.json
+++ b/config/config-mattermod.default.json
@@ -1,5 +1,5 @@
 {
-    "ListenAddress": "0.0.0.0:8086",
+    "ListenAddress": "0.0.0.0:8080",
     "MattermodURL": "",
     "GithubAccessToken": "",
     "GitHubTokenReserve": 10,
--- a/config/config-mattermod.default.json
+++ b/config/config-mattermod.default.json
@@ -120,7 +120,7 @@
     "MattermostWebhookURL": "",
     "MattermostWebhookFooter": "",
 
-    "MetricsServerPort": "8067",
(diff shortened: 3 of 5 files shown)


agnivade:
Excited for this!

Dockerfile
```diff
@@ -50,6 +50,6 @@ RUN for d in .ssh repos logs; do \
     done
 
 USER mattermod
-EXPOSE 8080
+EXPOSE 8080 9000
```
ethervoid, line 53: If we're going to use 9000 as the default port for metrics maybe we need to change the default port [here](https://github.com/mattermost/mattermost-mattermod/blob/master/config/config-mattermod.default.json#L123). What do you think?

khos2ow: That's true, I wanted to have a discussion around it before changing it. And to be clear that's going to be used on local development side and not the config being used in production.

agnivade: Doesn't matter really. It's up to the operator to configure whatever port they prefer, as long as it is changeable through the config.

khos2ow: And also on the similar topic [this](https://github.com/mattermost/mattermost-mattermod/blob/master/config/config-mattermod.default.json#L2) also has to change to `8080` as well.

khos2ow: @agnivade that's not entirely true, and not doable at the moment. The config, dynamically loaded into mattermod and start server on any arbitrary port is completely fine but on the underlying side of Dockerfile that said port have had to be exported already, otherwise there's no traffic gong in on that particular port for the server pick the traffic up.

ethervoid: Basically is to bring consistency to new people that come to the project. Having the metrics port in 8067 and the docker image using the 9000 could be confusing. Agree with the 8080 change too

khos2ow: Yes, I agree, there needs to be consistency. I personally prefer 9000 better because it's outside of known http ranges and known Prometheus ports are in 9000+ range. Otherwise we can just agree on any given port and use it everywhere for metrics server (as long as there's no conflct)

ethervoid: Works for me ;)

agnivade: We can use the value in the Dockerfile to set the config value as an environment variable; i.e. make the Dockerfile value dynamic as well.

I am fine with changing the default. But there are 6 separate places where the port is defined, so it's a bit problematic to maintain.

khos2ow: Although exposing dynamic port in Dockerfile based on environment variable is doable, but it's not recommended. And also one of the main reasons to originally put port in config file most probably was to control the port and prevent collision on a standalone test or deployment. But with moving to containerized version of deployment and also having Kubernetes to orchestrate it setting the port dynamically is not that beneficial anyways.

As you mentioned that port is in used in multiple places and it's completely fine from the point of view of Docker and K8s (i.e. no port collision). I mean if nobody in the projects uses standalone binary for testing and local development purposes we can just go ahead and hardcode the port within the go files anyways.

PR #211, opened 2020-07-31
Reduce replicas to 1.

metanerd:
#### Summary
Because of how the rate limit handling is done right now, we need to work with 1 replica.

### DIFF
--- a/deploy/overlays/prod/patches/deployment-replicas.yaml
+++ b/deploy/overlays/prod/patches/deployment-replicas.yaml
@@ -4,4 +4,4 @@ kind: Deployment
 metadata:
   name: mattermod
 spec:
-  replicas: 1
+  replicas: 3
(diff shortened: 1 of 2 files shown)


deploy/overlays/prod/patches/deployment-replicas.yaml
```diff
@@ -4,4 +4,4 @@ kind: Deployment
 metadata:
   name: mattermod
 spec:
-  replicas: 3
+  replicas: 1
```
khos2ow, line 7: Depending on the result of the ongoing discussion, if this meant to be a temporary workaround this change is completely fine. But if mattermod has to run only in one instance this file (`deployment-replicas.yaml`) and the corresponding line in `prod/kustomization.yaml` should be deleted.

PR #212, opened 2020-07-31
[MM-27431] Multiple metrics fixes

ethervoid:
#### Summary

This PR includes some fixes to the metrics:

- Fixed the missing registration of the githubCache metrics
- Fixed the order of parameters in github metrics that was wrong and was sending the method to the handler and other way around
- Reduce the path string to avoid too much unnecessary information as you can see in the next screenshot:

Screenshot from 2020-07-31 19-13-49

  now we plan to store "/repo/{user|org}/{repository}/{api_endpoint}" where api_endpoint could be something like: issues, branches, etc

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-27431

### DIFF
--- a/metrics/provider.go
+++ b/metrics/provider.go
@@ -123,6 +123,7 @@ func NewPrometheusProvider() *PrometheusProvider {
 		},
 		[]string{"method", "handler"},
 	)
+	provider.Registry.MustRegister(provider.githubCacheHits)
 
 	provider.githubCacheMisses = prometheus.NewCounterVec(
 		prometheus.CounterOpts{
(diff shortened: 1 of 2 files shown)


server/metrics.go
```diff
@@ -57,13 +58,16 @@ func (t *MetricsTransport) RoundTrip(req *http.Request) (resp *http.Response, er
 	if resp == nil && err != nil {
 		return resp, err
 	}
+	splittedPath := strings.Split(req.URL.Path, "/")
+	// This would leave path as "/repos/{user/org}/{repository}/issues"
+	path := strings.Join(splittedPath[:5], "/")
```
agnivade, line 63: Let's verify the length of the slice

metrics/provider.go
```diff
@@ -123,6 +123,7 @@ func NewPrometheusProvider() *PrometheusProvider {
 		},
 		[]string{"method", "handler"},
 	)
+	provider.Registry.MustRegister(provider.githubCacheHits)
```
agnivade, line 126: Ah I was wondering where these were gone.

PR #213, opened 2020-08-02
[MM-27399] Don't break for iteration if one PR fails

ethervoid:
#### Summary

Don't stop checking PR activity if one PR fails

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-27399


ethervoid:
Sure!

PR #214, opened 2020-08-04
use PR from event when checking if PR is mergeable

ashishbhate:
#### Summary

The production instance of mattermod seems to think that some merged and closed PRs have been closed but not merged. The closing of the PR triggers the event in mattermod, which then queries Github to check if the PR has been merged. The response from Github appears to be that the PR was not merged, although manually checking the API response from Github a little while later shows that the event has been merged. 

This leads me to suspect a caching issue. In this PR mattermod uses the merge state information from the event itself instead of querying Github again. According to the Github docs [1][2], the event contains all the information we need so we should not need to query Github again. 

[1] https://developer.github.com/webhooks/event-payloads/#pull_request 
[2] https://developer.github.com/v3/pulls/#get-a-pull-request

#### Ticket Link

- N/A, but I've made a follow-up issue to fix this issue across the code base: https://mattermost.atlassian.net/browse/MM-27487

### DIFF
--- a/server/cherry_pick.go
+++ b/server/cherry_pick.go
@@ -56,14 +56,14 @@ func (s *Server) checkIfNeedCherryPick(pr *model.PullRequest) {
 	ctx, cancel := context.WithTimeout(context.Background(), defaultRequestTimeout*time.Second)
 	defer cancel()
 
-	prCherryCandidate, _, err := s.GithubClient.PullRequests.Get(ctx, pr.RepoOwner, pr.RepoName, pr.Number)
-	if err != nil {
-		mlog.Error("Error getting the PR info", mlog.Err(err))
+	if !pr.Merged.Valid || !pr.Merged.Bool {
+		mlog.Info("PR not merged, not cherry picking", mlog.Int("PR Number", pr.Number), mlog.String("Repo", pr.RepoName))
 		return
 	}
 
-	if !prCherryCandidate.GetMerged() {
-		mlog.Info("PR not merged, not cherry picking", mlog.Int("PR Number", prCherryCandidate.GetNumber()), mlog.String("Repo", pr.RepoName))
+	prCherryCandidate, _, err := s.GithubClient.PullRequests.Get(ctx, pr.RepoOwner, pr.RepoName, pr.Number)
+	if err != nil {
+		mlog.Error("Error getting the PR info", mlog.Err(err))
 		return
 	}
 


server/cherry_pick.go
```diff
@@ -56,14 +56,14 @@ func (s *Server) checkIfNeedCherryPick(pr *model.PullRequest) {
...
-	if err != nil {
-		mlog.Error("Error getting the PR info", mlog.Err(err))
+	if !pr.Merged.Valid || !pr.Merged.Bool {
+		mlog.Info("PR not merged, not cherry picking", mlog.Int("PR Number", pr.Number), mlog.String("Repo", pr.RepoName))
 		return
 	}
 
-	if !prCherryCandidate.GetMerged() {
-		mlog.Info("PR not merged, not cherry picking", mlog.Int("PR Number", prCherryCandidate.GetNumber()), mlog.String("Repo", pr.RepoName))
+	prCherryCandidate, _, err := s.GithubClient.PullRequests.Get(ctx, pr.RepoOwner, pr.RepoName, pr.Number)
```
agnivade, line 64: We need to think about the disparity between model.PullRequest and github.PullRequest a bit more. Just because that field is not there, we have to query GH again and waste those tokens.

server/cherry_pick.go
```diff
@@ -56,14 +56,14 @@ func (s *Server) checkIfNeedCherryPick(pr *model.PullRequest) {
 	ctx, cancel := context.WithTimeout(context.Background(), defaultRequestTimeout*time.Second)
 	defer cancel()
 
-	prCherryCandidate, _, err := s.GithubClient.PullRequests.Get(ctx, pr.RepoOwner, pr.RepoName, pr.Number)
-	if err != nil {
-		mlog.Error("Error getting the PR info", mlog.Err(err))
+	if !pr.Merged.Valid || !pr.Merged.Bool {
```
agnivade, line 59: All that SQL abstraction is totally lost now :D I am wondering if a helper method like IsMerged to abstract away this `NullBool` logic makes sense?

server/cherry_pick.go
cpanato: +1

ashishbhate: I opened this PR to fix the immediate issue we were seeing in the logs.
 
The follow-up ticket (https://mattermost.atlassian.net/browse/MM-27487) tracks the effort of updating all the models, and bringing this change to the other objects.

agnivade: Oh it was a general observation, and not meant to be done in this ticket. Sorry if it implied otherwise.

server/cherry_pick.go
ashishbhate: The extremist in me says lets just drop the DB ;)

The go-github library has getters like `pullRequest.GetMerged()`. We could model our model on their model :)

server/cherry_pick.go
ashishbhate: >Oh it was a general observation, and not meant to be done in this ticket. Sorry if it implied otherwise.

No worries. I didn't infer it that way :) 

I've linked this PR in the Jira ticket, so the lucky dev fixing that issue has context.

PR #215, opened 2020-08-04
blocklist: refactor block list paths to accept a pattern and rename the config

cpanato:
#### Summary
Juho is reviewing the CI configs and we need to add more files in the block list, today we need to set the names in the config and there is no option to set a pattern like `**/*.sh` for example

This PR address that adding an option to add patterns to check the files.
Also, we rename the variable from blacklist to blocklist.

After we merge this we need to update the config to use `BlocklistPaths`

Added tests as well

Feedbacks are welcome

#### Ticket Link
n/a


metanerd:
Adding @agnivade instead of @jespino because Jesus is OoO.

server/circleci.go
```diff
@@ -49,15 +51,16 @@ func (s *Server) triggerCircleCiIfNeeded(ctx context.Context, pr *model.PullRequ
...
-		for _, blockListPath := range s.Config.BlacklistPaths {
-			if prFile.GetFilename() == blockListPath {
-				mlog.Error("File is on the blocklist and will not retrigger circleci to give the contexts", mlog.String("repo", pr.RepoName), mlog.Int("pr", pr.Number), mlog.String("Fullname", pr.FullName))
-				msg := fmt.Sprintf("The file `%s` is in the blocklist and should not be modified from external contributors, please if you are part of the Mattermost Org submit this PR in the upstream.\n /cc @mattermost/core-security @mattermost/core-build-engineers", prFile.GetFilename())
-				s.sendGitHubComment(ctx, pr.RepoOwner, pr.RepoName, pr.Number, msg)
-				return
-			}
+	blockMessage, err := s.validateBlockPaths(prFiles)
+	if err != nil {
+		if blockMessage != "" {
```
agnivade, line 56: I think we can improve the error handling a bit further by having a separate error type for blockList error and not returning any `blockMessage` at all. What we are trying to do here is check whether there is an error or not, then differentiate whether it's a blocklist error or not by inspecting the `blockMessage` variable.

IMO, if we have a generic error for the filepath failure case, and a separate error type for an actual blockList error, we can handle this case more elegantly.

server/circleci_test.go
```diff
@@ -37,3 +38,100 @@ func TestAreNotAllExpectedJobs(t *testing.T) {
...
+				},
+				{
+					Filename: github.String(".circleci/honk.test"),
+				},
+				{
+					Filename: github.String("build/validone-honk.go"),
+				},
+			},
+			expectedError:        true,
+			expectedMessageError: "The file `.circleci/config.yml` is in the blocklist and should not be modified from external contributors, please if you are part of the Mattermost Org submit this PR in the upstream.\n /cc @mattermost/core-security @mattermost/core-build-engineers",
```
agnivade, line 63: Comparing error strings is not very ideal. I think the objective here is to compare the filenames, and not the entire error message. If we have a custom error type as I mentioned above, we can solve this in a cleaner way.

server/circleci.go
cpanato: did some changes after our chat. PTAL
I think that maybe still not in the way you want, but was what I understood :)

server/circleci_test.go
cpanato: change that @agnivade

config/config-mattermod.default.json
```diff
@@ -82,7 +82,7 @@
     "CLAGithubStatusContext": "",
     "SignedCLAURL": "",
     "NeedsToSignCLAMessage": "",
-    "BlacklistPaths": [],
+    "BlocklistPaths": [],
```
hanzei, line 85: Should be kept backwards compatible and we do the cleanup in https://github.com/mattermost/mattermost-mattermod/pull/144?

server/circleci.go
```diff
@@ -104,6 +103,57 @@ type PipelineTriggeredResponse struct {
...
+// Error implements the error interface.
+func (e *BlockPathValidationError) Error() string {
+	return "files in the Block List " + strings.Join(e.files, ",")
+}
+
+func (e *BlockPathValidationError) BlockListFiles() []string {
+	return e.files
+}
+
+// ReportBlockFiles implements the error interface.
```
agnivade, line 119: Copy-paste error :)

cpanato: done

config/config-mattermod.default.json
cpanato: I can rename to `BlockListPathsGlobal`

cpanato:
@metanerd can you schedule this to be merged and changed? if you need help let me know

metanerd:
Deployed the change. Thank you!

PR #216, opened 2020-08-05
MM-27435: Split into a separate job server

agnivade:
This PR creates a separate service for the job server
and removes the functionality from the mattermod server.

A separate docker image with its own Dockerfile is provided.

The same config file needs to be used for the new service.
It is preferred that a separate GitHub token is used to separate
the services better in case one token is deleted or expired, it should
not affect the other service.


agnivade:
I have left the updates to the deploy directory to someone with more devops knowledge.

cmd/jobserver/main.go
```diff
@@ -0,0 +1,105 @@
...
+		mlog.Error("Unable to load Job Server config", mlog.Err(err), mlog.String("file", configFile))
+		os.Exit(1)
+	}
+	if err = server.SetupLogging(config); err != nil {
+		mlog.Error("Unable to configure logging", mlog.Err(err))
+		os.Exit(1)
+	}
+
+	// Metrics system
+	metricsProvider := metrics.NewPrometheusProvider()
```
ethervoid, line 37: To get this metrics we have to include this service as target in Prometheus // @khos2ow

agnivade: Oh right, good catch. Let me update the Dockerfile.

agnivade:
Good point, I think for now we can use different copies of the config files. In the future, we can look to viper-ize this to override individual settings with environment variables.

Yeah in production, we already used to run 3 replicas of mattermod, all of them listening on 9000. So that won't be an issue.

Makefile
```diff
@@ -52,22 +55,32 @@ test:
 build: clean
 	@echo Building
 	$(GO) build -o dist/mattermod ./cmd/mattermost-mattermod
+	$(GO) build -o dist/jobserver ./cmd/jobserver
```
khos2ow, line 58: In order to reduce build time (specially on release) may be we can separate these into three targets:

```
build: build-mattermod build-jobserver

build-mattermod:
    $(GO) build ....

build-jobserver:
    $(GO) build ....
```

Dockerfile.jobserver
```diff
@@ -0,0 +1,47 @@
...
+RUN apk add --update --no-cache ca-certificates bash make gcc musl-dev git openssh wget curl
+
+WORKDIR /go/src/mattermod
+
+COPY go.mod .
+COPY go.sum .
+RUN go mod download
+
+COPY . .
+RUN CGO_ENABLED=0 make build
```
khos2ow, line 12: See my comment below for renaming, if you change it there this also needs to change to `make build-jobserver`. and the same line in `Dockerfile` needs to change to `make build-mattermod`

cmd/jobserver/main.go
khos2ow: That's true, and will be part of the deployment of this new jobserver.

khos2ow:
@metanerd @jaydeland we decided to use the same PR for the feature and the deployment. The go implementation review is done (you're more than welcome to take a look again too) but the rest (Makefile, CircleCI config and `deploy/` manifests) need a fresh review.

agnivade:
A small note that now while running mattermod with replicas, we might want to reduce the token reserve for each replica. It's not necessary to do it upfront, but something to keep an eye on metrics if we see rate limit issues in production. @ethervoid can provide some more guidance into what a better value can be.

agnivade:
The config file is the same. I'll make a note to add a section on how to start the job server once this gets merged.

PR #217, opened 2020-08-05
MM-27192: move hardcoded filesystem paths to config

ashishbhate:
#### Summary

- Move hardcoded paths to configuration. 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-27192


agnivade:
@khos2ow had some concerns on this. I will wait till I see a green signal from him.

server/cherry_pick.go
```diff
@@ -106,9 +107,20 @@ func (s *Server) doCherryPick(ctx context.Context, version string, milestoneNumb
...
+
+	if s.Config.RepoFolder == "" {
+		return "", errors.Errorf("path to folder containing local checkout of repositories is not set in the config")
+	}
+	repoFolder := filepath.Join(s.Config.RepoFolder, pr.RepoName)
+
+	if s.Config.ScriptsFolder == "" {
+		return "", errors.Errorf("path to folder containing the cherry-pick.sh script is not set in the config")
+	}
+	cherryPickScript := filepath.Join(s.Config.ScriptsFolder, "cherry-pick.sh")
```
ashishbhate, line 119: Maybe the empty checks should be moved to the config loader?

ashishbhate:
>@khos2ow had some concerns on this. I will wait till I see a green signal from him.

I don't think leads to much disruption in our deployment story. Should be just the addition of these two new fields in our deployment mattermod config,

 Independent of our production deployment, this makes running mattermod on a local dev machine much cleaner.

ashishbhate:
>@khos2ow had some concerns on this. I will wait till I see a green signal from him.

BTW, he has already been added as a reviewer, I wont merge unless everyone approves, so this shouldn't be a blocker for you to add your own thoughts and review.

ashishbhate:
I understand from a production POV this is not necessary, but is it a blocker?

I think it is necessary for running local binaries because currently I have to maintain an out of tree patch for running mattermod on my local machine, just because there are arbitrary paths hardcoded in the codebase. 

>1. build time with docker:
>2. run time with docker image:

If its difficult to keep `ScriptsFolder` in sync, how about adding `/app/scripts/` as a default script path in the config loader? Similar to how the the config file is searched for in certain fixed locations - https://github.com/mattermost/mattermost-mattermod/blob/master/server/config.go#L175-L187

This will not require any deployment changes, because leaving it empty is the same as setting it to `/app/scripts`, while still giving the dev the flexibility of adjusting the options for his local setup.

khos2ow:
No it's not blocker at all, just a bit redundant.

> I think it is necessary for running local binaries

Technically you don't need to run the binaries locally and you can use local tiny Kubernetes cluster for your local development too. But I understand if you don't want to :)

server/cherry_pick.go
agnivade: This is a good candidate for @isacikgoz's config validation library. Low priority, but good to have.

ashishbhate:
>Technically you don't need to run the binaries locally and you can use local tiny Kubernetes cluster for your local development too. But I understand if you don't want to :)

The only issue I have with running the binaries through docker / kubernetes is that I have to jump through a few hoops to get my local config file available (through some mount) in the containers. 
Running binaries is just a single simple bash command.

ashishbhate:
Summary
- Move hardcoded paths to configuration.

Ticket Link
- https://mattermost.atlassian.net/browse/MM-27192

PR #220, opened 2020-08-06
[MM-27019] Migrate PR handler and add some tests

isacikgoz:
#### Summary
This PR migrates `pullRequestEventHandler` and add some testing around data validation but specific tests for events are not included.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27019


server/pull_request_handler.go
```diff
@@ -157,29 +181,45 @@ func (s *Server) handlePullRequestEvent(ctx context.Context, event *PullRequestE
 		}
 	}
 
-	s.checkPullRequestForChanges(ctx, pr)
+	changed, err := s.checkPullRequestForChanges(ctx, pr)
+	if err != nil {
+		mlog.Error("Could not check changes for PR", mlog.Err(err))
```
ethervoid, line 186: Maybe we should return 400 or 500 error here

server/pull_request_handler.go
```diff
@@ -157,29 +181,45 @@ func (s *Server) handlePullRequestEvent(ctx context.Context, event *PullRequestE
...
 		}
 
-		return
+		return false, nil
 	}
 
-	prHasChanges := false
+	var prHasChanges bool
 
 	for _, label := range pr.Labels {
```
ethervoid, line 224: Maybe we can extract this part of the code and create a new function that receives two string slices and returns the diff between them so we can reduce the amount of boilerplate here. What do you think?

agnivade:
Sorry this will get a bit delayed on review.

server/pull_request_handler.go
isacikgoz: Makes sense, I also thought to return `500` for this but didn't want to introduce behavioral change. But I'm open to change it 👍

ethervoid: I'll do it. All the HTTP errors are new but I think is necessary to be aware of problems. Those errors would be reflected in the metrics so it's a good thing

server/pull_request_handler.go
```diff
@@ -248,13 +292,15 @@ func (s *Server) checkPullRequestForChanges(ctx context.Context, pr *model.PullR
 	if prHasChanges {
 		mlog.Info("pr has changes", mlog.Int("pr", pr.Number))
```
agnivade, line 293: We should remove this. The logging is already happening at the upper level.

server/pull_request_handler.go
isacikgoz: Yeah, makes sense to avoid repetition.

PR #221, opened 2020-08-06
Included rate limit errors metrics

ethervoid:
#### Summary

This PR includes rate limit errors metrics

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-27549


server/rate_limit.go
```diff
@@ -14,10 +14,12 @@ import (
...
 	base    http.RoundTripper
+	metrics MetricsProvider
 }
 
 func (t *RateLimitTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 	if err := t.limiter.Wait(req.Context()); err != nil {
+		t.metrics.IncreaseRateLimiterErrors()
 		return nil, err
 	}
 	return t.base.RoundTrip(req)
```
agnivade, line 25: I would also like to add a responsewriter middleware which checks for 403 responses and increments ratelimit errors in case we get a 403 from GitHub. This is because our token reserve is larger than the allowed rate limits, so in case we _think_ that we are within limits, but we make a lot of POST calls, the limiter would allow them, and trigger a rate limit error.

server/metrics.go
```diff
@@ -54,26 +61,53 @@ func (t *MetricsTransport) RoundTrip(req *http.Request) (resp *http.Response, er
...
 	}
 
-	return resp, err
+	if resp.Body != nil {
+		msg := struct {
+			Message          string `json:"message"`
+			DocumentationURL string `json:"documentation_url"`
+		}{}
+		// Read body to check if there is an error message,
+		// then close it and reassing the body again for the
```
agnivade, line 101: re-assigning

server/metrics.go
```diff
@@ -54,26 +61,53 @@ func (t *MetricsTransport) RoundTrip(req *http.Request) (resp *http.Response, er
...
+		// next layer so it could read the body without problem
+		bodyBytes, _ := ioutil.ReadAll(resp.Body)
```
agnivade, line 103: Any reason we are ignoring the error here? Seems to me that we can easily return an error and propagate it upwards.

server/metrics.go
```diff
@@ -54,26 +61,53 @@ func (t *MetricsTransport) RoundTrip(req *http.Request) (resp *http.Response, er
...
+		resp.Body.Close()
+		err := json.Unmarshal(bodyBytes, &msg)
+		if err == nil && statusCode == "403" && strings.Contains(msg.Message, "temporarily blocked") {
```
agnivade, line 106: Same here too

server/github_test.go
```diff
@@ -283,6 +283,7 @@ func TestRateLimitTransport(t *testing.T) {
 		metricsMock.EXPECT().ObserveGithubRequestDuration(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).AnyTimes()
 		metricsMock.EXPECT().IncreaseGithubCacheMisses(gomock.Any(), gomock.Any()).AnyTimes()
 		metricsMock.EXPECT().IncreaseGithubCacheHits(gomock.Any(), gomock.Any()).AnyTimes()
+		metricsMock.EXPECT().IncreaseRateLimiterErrors().AnyTimes()
```
agnivade, line 286: It should be possible to add a test that returns a 403 with the correct error message and confirm that the metric is being incremented exactly once to have a stricter check.

PR #222, opened 2020-08-10
[MM-26783] server/update_branch: tweak errors and messages

isacikgoz:
#### Summary
I ended up with a slightly different approach I guess :) But we won't lose the actual error messages while passing errors to upper level :p 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26783

### DIFF
--- a/server/update_branch.go
+++ b/server/update_branch.go
@@ -2,6 +2,7 @@ package server
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


server/update_branch_test.go
```diff
@@ -51,38 +51,38 @@ func TestHandeUpdateBranch(t *testing.T) {
 	s.GithubClient.Issues = is
 
 	t.Run("random user", func(t *testing.T) {
-		*msg = MsgCommenterPermission
+		*msg = msgCommenterPermission
 
 		err := s.handleUpdateBranch(ctx, "someone", pr)
 		require.Error(t, err)
-		require.EqualError(t, ErrCommenterPermission, err.Error())
+		require.EqualError(t, err, "commenter does not have permissions")
```
agnivade, line 58: Wondering if it would be better if we matched the `source` field with *msg. That's a more cleaner way without the string matching.

server/update_branch.go
```diff
@@ -11,40 +10,48 @@ import (
...
+		return "could not update pull request"
+	default:
+		panic("unhandled error type")
+	}
+}
+
 func (s *Server) handleUpdateBranch(ctx context.Context, commenter string, pr *model.PullRequest) error {
-	var err error
-	var msg string
+	var e *updateError
```
agnivade, line 37: Nit: `e` as a single letter is too short a variable name for the entire method. 0/5 on the new name, but a longer name would be more idiomatic IMO.

server/update_branch.go
```diff
@@ -54,15 +61,13 @@ func (s *Server) handleUpdateBranch(ctx context.Context, commenter string, pr *m
...
-		msg = MsgUpdatePullRequest
-		err = ErrUpdatePullRequest
-		return err
+		e = &updateError{msgUpdatePullRequest}
+		return e
 	}
 	if err != nil && !strings.Contains("job scheduled on GitHub side; try again later", err.Error()) {
-		msg = MsgUpdatePullRequest
+		e = &updateError{msgUpdatePullRequest}
 		return err
```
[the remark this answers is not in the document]
isacikgoz: We shouldn't lose the err information there, I `err` with `e` instead.

server/update_branch.go
isacikgoz: The first longer name came into my mind was `ue` 😝

PR #226, opened 2020-08-13
[MM-27263] server/cherry_pick: queue up cherry pick requests

isacikgoz:
#### Summary
This PR adds ability to queue up cherry-pick requests. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27263


server/cherry_pick.go
```diff
@@ -23,6 +23,47 @@ import (
...
+	version   string
+	milestone int
+}
+
+func (s *Server) listenCherryPickRequests() {
+	for {
+		select {
+		case job := <-s.cherryPickRequests:
+			ctx, cancel := context.WithTimeout(context.Background(), defaultRequestTimeout*time.Second)
+			defer cancel()
```
agnivade, line 42: This will never execute. We need to run this entire thing inside a closure.

server/cherry_pick.go
```diff
@@ -23,6 +23,47 @@ import (
...
+			pr := job.pr
+			cmdOut, err := s.doCherryPick(ctx, strings.TrimSpace(job.version), &job.milestone, pr)
+			if err != nil {
+				msg := fmt.Sprintf("Error trying doing the automated Cherry picking. Please do this manually\n\n```\n%s\n```\n", cmdOut)
+				s.sendGitHubComment(ctx, pr.RepoOwner, pr.RepoName, pr.Number, msg)
+				mlog.Error("Error while cherry picking", mlog.Err(err))
+			}
+		case <-s.stopChan:
+			close(s.cherryPickRequests)
+			s.drainCherryPickTasks()
```
agnivade, line 52: Not sure if we need such a complicated way of doing things. We can just range over the channel, and then close it and send another signal either through waitgroup or another channel to the parent goroutine. I would also prefer the channels to be prefixed with the name `cherryPick` so that it's clear it's for cherry picking and not meant to be used at a server level.

```go
func (s *Server) listenCherryPickRequests() {
 defer func() {
   close(s.cherryPickStoppedChan)
   // or alternatively use a waitgroup. Either is fine.
 }()

 for job := range s.cherryPickRequests {
  // handle job
 }
}
```

```go
func (s *Server) stopCherryPickListener() {
 close(c.cherryPickRequests)
 <-s.cherryPickStoppedChan
}
```
Then we just have a single place to do everything. And the pending jobs in the queue don't get canceled when the server is shutting down.

server/cherry_pick.go
```diff
@@ -41,10 +82,18 @@ func (s *Server) handleCherryPick(ctx context.Context, commenter, body string, p
 	if !pr.Merged.Valid || !pr.Merged.Bool {
 		return nil
 	}
-	cmdOut, err := s.doCherryPick(ctx, strings.TrimSpace(args[1]), nil, pr)
-	if err != nil {
-		msg = fmt.Sprintf("Error trying doing the automated Cherry picking. Please do this manually\n\n```\n%s\n```\n", cmdOut)
-		return err
+
+	select {
+	case <-s.stopChan:
```
agnivade, line 87: Not sure if we need to quit from here. See my above comment.

server/cherry_pick.go
isacikgoz: That was my initial thought about this but in that case we need to wait every cherry-pick request to complete. If there are too many requests queued up, server shutdown will take too much time, I did this to be able to cancel remaining requests.

agnivade: I see. But in your case, we are cancelling everything the moment we get a stop request. I think it is reasonable for the server to have to wait for a while to flush out pending cherry-pick requests, within a timeout of course.

I see 2 options here:

- How much of a time on average does a cherry-pick take? Depending on that, we might want to reduce the buffer size. So then server shutdown never takes too long.
- We can add additional timeout mechanism in the `stopCherryPickListener` method to exit the server if it has to wait beyond a certain time. But then those "cancelling requests" comment doesn't get sent out.

I think reducing the buffer size is an automatic way of ensuring the server shuts down in a reasonable time and also ensuring that the user gets a response.

isacikgoz: I think we can prioritize informing user about the outcome if it is canceled/failed/succeeded. I think the answer of "how often do we restart the server" can make more sense to give a decision about this. Also, cherry-pick shouldn't take too much but it depends to too many components. Sometimes it takes too long to do `gh checkout pr n`. So the option 1 makes more sense to me.

From my understanding we have to decide to give-up cancelling requests over code complexity. Anyway, we can reduce the queue size at the beginning and wait every cherry-pick request to complete. After some usage we may have a better picture what path should we use.

agnivade: > From my understanding we have to decide to give-up cancelling requests over code complexity. 

I don't mind the code complexity really. IMO, this is what should be done: while closing the server, we send a signal to the goroutine, and wait for x seconds for it to return. In those x seconds, it can flush out the queue. And when the timeout is over, the remaining items need to be canceled and sent to the user.

This is what I would want in an ideal world. This has almost everything in your first solution, but instead of outright cancelling, it waits for a certain period. Now I think implementing this can get quite complicated. But I am 0/5 if you see a reasonable way to do this.

If not, I think our current solution is good enough :) So yeah, we are compromising on code complexity for our "ideal" solution. But if the "ideal" solution is doable, then I'm all for it.

server/cherry_pick.go
```diff
@@ -41,10 +86,16 @@ func (s *Server) handleCherryPick(ctx context.Context, commenter, body string, p
...
 	}
-	cmdOut, err := s.doCherryPick(ctx, strings.TrimSpace(args[1]), nil, pr)
-	if err != nil {
-		msg = fmt.Sprintf("Error trying doing the automated Cherry picking. Please do this manually\n\n```\n%s\n```\n", cmdOut)
-		return err
+
+	select {
+	case s.cherryPickRequests <- &cherryPickRequest{
+		pr:      pr,
+		version: strings.TrimSpace(args[1]),
```
ashishbhate, line 93: I think we need to change the type of `milestone` to `*int` 

In `doCherryPick` on line 207, we check if `milestoneNumber != nil`, which is this case it would not be nil because milestone would be 0.

server/cherry_pick.go
```diff
@@ -119,12 +176,13 @@ func (s *Server) doCherryPick(ctx context.Context, version string, milestoneNumb
 	cherryPickScript := filepath.Join(s.Config.ScriptsFolder, "cherry-pick.sh")
 
 	releaseBranch := fmt.Sprintf("upstream/%s", version)
-
+	fmt.Println(strings.Join([]string{cherryPickScript, releaseBranch, strconv.Itoa(pr.Number), pr.MergeCommitSHA}, " "))
```
ashishbhate, line 179: Missed print or maybe change to mlog.

server/cherry_pick.go
```diff
@@ -23,6 +23,51 @@ import (
...
+				msg := fmt.Sprintf("Error trying doing the automated Cherry picking. Please do this manually\n\n```\n%s\n```\n", cmdOut)
+				s.sendGitHubComment(ctx, pr.RepoOwner, pr.RepoName, pr.Number, msg)
+				mlog.Error("Error while cherry picking", mlog.Err(err))
+			}
+		}()
+	}
+}
+
+func (s *Server) finishCherryPickRequests() {
+	close(s.cherryPickRequests)
```
ashishbhate, line 58: `finishCherryPickRequests` is the first thing that's called when `server.Stop` is called.
The first thing we do here is close the channel.

Isn't it possible for a request/event to come in while we're in server shutdown, but still in this function? This would cause the server to still call `handleCherryPick` or `checkIfNeedCherryPick` which writes to this channel which would lead to a panic.

server/cherry_pick.go
```diff
@@ -41,10 +86,16 @@ func (s *Server) handleCherryPick(ctx context.Context, commenter, body string, p
 	if !pr.Merged.Valid || !pr.Merged.Bool {
 		return nil
 	}
-	cmdOut, err := s.doCherryPick(ctx, strings.TrimSpace(args[1]), nil, pr)
-	if err != nil {
-		msg = fmt.Sprintf("Error trying doing the automated Cherry picking. Please do this manually\n\n```\n%s\n```\n", cmdOut)
-		return err
+
+	select {
```
agnivade, line 90: My apologies, we do need the case for closing of the channel. Otherwise it will panic once we close the channel. Sorry I forgot about that case earlier.

server/cherry_pick.go
```diff
@@ -23,6 +23,51 @@ import (
...
+	}
+}
+
+func (s *Server) finishCherryPickRequests() {
+	close(s.cherryPickRequests)
+	select {
+	case <-time.After(defaultRequestTimeout * time.Second):
+		ctx, cancel := context.WithTimeout(context.Background(), defaultRequestTimeout*time.Second)
+		defer cancel()
+		for job := range s.cherryPickRequests {
```
agnivade, line 63: Hmm .. this is okay, but it races with the original goroutine. So what we are doing is essentially speeding up the shutdown. That's okay I guess. But let's add a comment to explain what is going on here.

server/cherry_pick.go
isacikgoz: Good catch 👍 Reverted it something similar to the my original proposal :)

server/cherry_pick.go
isacikgoz: No worries, I got confused as well.

server/cherry_pick.go
```diff
@@ -23,6 +23,54 @@ import (
...
+	close(s.cherryPickStopChan)
+	close(s.cherryPickRequests)
+	select {
+	case <-time.After(5 * time.Second):
+		ctx, cancel := context.WithTimeout(context.Background(), defaultRequestTimeout*time.Second)
+		defer cancel()
+		// While consuming requests here, listenCherryPickRequests routine will continue
+		// to listen as well. We're just trying to cancel requests as much as we can.
+		for job := range s.cherryPickRequests {
+			msg := "Cherry picking is canceled due to server shutdown."
```
agnivade, line 67: Nit: we can move this above the for loop.

server/cherry_pick.go
```diff
@@ -41,10 +89,18 @@ func (s *Server) handleCherryPick(ctx context.Context, commenter, body string, p
...
 	}
-	cmdOut, err := s.doCherryPick(ctx, strings.TrimSpace(args[1]), nil, pr)
-	if err != nil {
-		msg = fmt.Sprintf("Error trying doing the automated Cherry picking. Please do this manually\n\n```\n%s\n```\n", cmdOut)
-		return err
+
+	select {
+	case <-s.cherryPickStopChan:
+		return errors.New("server is closing")
+	case s.cherryPickRequests <- &cherryPickRequest{
```
ashishbhate, line 96: When the server is in shutdown, both of these channels will be closed.  The go runtime considers a write to a closed channel as a valid condition to proceed in a select.

Since select will choose one of these cases at random, this will randomly panic if an event comes in when we're in shutdown.

server/server.go
```diff
@@ -31,17 +31,20 @@ import (
...
+	CircleCiClientV2      CircleCIService
+	OrgMembers            []string
+	Builds                buildsInterface
+	commentLock           sync.Mutex
+	StartTime             time.Time
+	awsSession            *session.Session
+	Metrics               MetricsProvider
+	cherryPickRequests    chan *cherryPickRequest
+	cherryPickStopChan    chan struct{}
+	cherryPickStoppedChan chan struct{}
```
ashishbhate, line 47: -100/5 nit, since the other channels have a "chan" suffix, then I think `cherryPickRequests` should be `cherryPickRequestsChan` too

server/cherry_pick.go
agnivade: Yeah that's true. It's my fault really, I just somehow ignored it, when I shouldn't have. Thanks for digging into it..

I think we can do a non-blocking read of the stop channel to check if the channel is closed or not. And then move to the next case.

isacikgoz: Thanks for digging into this @ashishbhate I really appreciate it 🤗 I quite rushed into this to complete before my time off. Eventually it remained after it, but better late than buggy :)

server/server.go
isacikgoz: I think `cherryPickRequests` fits nice on the purpose, but the other cases `stop` and `stopped` seem to be quite dull without a `chan` suffix.

ashishbhate (approved):
Looks great! :100:

PR #227, opened 2020-08-20
MM-27986: Fix error in reading non-403 responses

agnivade:
This PR fixes 2 issues.
1. We were always trying to read the response body irrespective of the status code.
This was unnecessary because a 200 response will not contain the response that we were expecting.
This would cause errors like "json: cannot unmarshal array into Go value of type struct"

2. In case of a failure, we were not resetting the body to be read further downstream which would
lead to errors like "http: read on closed response body"


ethervoid (approved):
LTGM! Thx for the fix

PR #228, opened 2020-08-20
Add logs and increase default cron task timeouts

ethervoid:
#### Summary

This PR adds two things:

- Debug logs for GH requests
- Increase the default timeout for cron tasks. Looks like we're getting timeouts since we added the pagination to get all the results

### DIFF
--- a/server/enterprise.go
+++ b/server/enterprise.go
@@ -26,7 +26,7 @@ func (s *Server) triggerEnterpriseTests(pr *model.PullRequest) {
 
 	triggerInfo, err := s.getPRInfo(ctx, pr)
 	if err != nil {
-		mlog.Debug("error trying to get pr info", mlog.Err(err))
+		mlog.Warn("error trying to get pr info", mlog.Err(err))
 		s.createEnterpriseTestsErrorStatus(ctx, pr, err)
 		return
 	}
--- a/server/enterprise.go
+++ b/server/enterprise.go
@@ -3,6 +3,7 @@ package server
 import (
 	"context"
 	"errors"
+	"fmt"
 	"net/http"
 	"regexp"
 	"strconv"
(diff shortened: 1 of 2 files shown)


server/enterprise.go
```diff
@@ -59,6 +59,11 @@ type EETriggerInfo struct {
 func (s *Server) getPRInfo(ctx context.Context, pr *model.PullRequest) (info *EETriggerInfo, err error) {
 	pullRequest, _, err := s.GithubClient.PullRequests.Get(ctx, s.Config.Org, pr.RepoName, pr.Number)
 	if err != nil {
+		mlog.Debug("Error trying to get pr",
```
agnivade, line 62: I would prefer it to be wrapped with a context string here, and then logged at the call site.

server/enterprise.go
```diff
@@ -105,6 +110,11 @@ func (s *Server) getBranchWithSameName(ctx context.Context, remote string, repo
 	ghBranch, r, err := s.GithubClient.Repositories.GetBranch(ctx, remote, repo, ref)
 	if err != nil {
 		if r == nil || r.StatusCode != http.StatusNotFound {
+			mlog.Debug("Error trying to get branch",
```
agnivade, line 113: Same as above. Add context and propagate upwards.

ethervoid:
@agnivade I've added some changes I hope that's what you're refering too :sweat_smile:

server/enterprise.go
```diff
@@ -25,6 +26,7 @@ func (s *Server) triggerEnterpriseTests(pr *model.PullRequest) {
 
 	triggerInfo, err := s.getPRInfo(ctx, pr)
 	if err != nil {
+		mlog.Debug("error trying to get pr info", mlog.Err(err))
```
agnivade, line 29: mlog.Warn would be preferred.

agnivade (approved):
Yep, this is what I wanted. Proper error propagation with context.

server/enterprise.go
ethervoid: :+1: I didn't want to clutter the logs with it but you're right better warn

PR #233, opened 2020-09-09
Clone git repos if they are not present locally

agnivade:
- If the repo folder is not present locally, we clone it
and then set the upstream remote.
- While here, we also return error from returnToMaster function.
- And also, use CombinedOutput instead of Output which contains
output from standard error also. This can be helpful to debug errors.

https://mattermost.atlassian.net/browse/MM-27264

### DIFF
--- a/server/cherry_pick.go
+++ b/server/cherry_pick.go
@@ -341,24 +341,20 @@ func cloneRepo(ctx context.Context, dir, upstreamSlug, originSlug, repoName stri
 	cmd := exec.CommandContext(ctx, "git", "config", "user.name")
 	if out, err := runCommandWithOutput(cmd, dir); err != nil {
 		return err
-	} else {
-		if out == "" { // this means username is not set
-			cmd = exec.CommandContext(ctx, "git", "config", "--global", "user.name", "mattermost-build")
-			if err = runCommand(cmd, dir); err != nil {
-				return err
-			}
+	} else if out == "" { // this means username is not set
+		cmd = exec.CommandContext(ctx, "git", "config", "--global", "user.name", "mattermost-build")
+		if err = runCommand(cmd, dir); err != nil {
+			return err
 		}
 	}
 
 	cmd = exec.CommandContext(ctx, "git", "config", "user.email")
 	if out, err := runCommandWithOutput(cmd, dir); err != nil {
 		return err
-	} else {
-		if out == "" { // this means email is not set
-			cmd = exec.CommandContext(ctx, "git", "config", "--global", "user.email", "build@mattermost.com")
-			if err = runCommand(cmd, dir); err != nil {
-				return err
-			}
+	} else if out == "" { // this means email is not set
+		cmd = exec.CommandContext(ctx, "git", "config", "--global", "user.email", "build@mattermost.com")
+		if err = runCommand(cmd, dir); err != nil {
+			return err
 		}
 	}
 
--- a/server/cherry_pick.go
+++ b/server/cherry_pick.go
@@ -211,7 +211,6 @@ func (s *Server) doCherryPick(ctx context.Context, version string, milestoneNumb
 	cmd.Env = append(
 		os.Environ(),
 		os.Getenv("PATH"),
-		fmt.Sprintf("DRY_RUN=%t", false),
 		fmt.Sprintf("ORIGINAL_AUTHOR=%s", pr.Username),
 		fmt.Sprintf("GITHUB_USER=%s", s.Config.GithubUsername),
 		fmt.Sprintf("GITHUB_TOKEN=%s", s.Config.GithubAccessTokenCherryPick),
--- a/server/cherry_pick.go
+++ b/server/cherry_pick.go
@@ -185,8 +185,8 @@ func (s *Server) doCherryPick(ctx context.Context, version string, milestoneNumb
 	}
 	repoFolder := filepath.Join(s.Config.RepoFolder, pr.RepoName)
 
-	if _, err := os.Stat(repoFolder); os.IsNotExist(err) {
-		err := cloneRepo(s.Config.RepoFolder, s.Config.Org+"/"+pr.RepoName, s.Config.GithubUsername+"/"+pr.RepoName, pr.RepoName)
+	if _, err = os.Stat(repoFolder); os.IsNotExist(err) {
+		err = cloneRepo(s.Config.RepoFolder, s.Config.Org+"/"+pr.RepoName, s.Config.GithubUsername+"/"+pr.RepoName, pr.RepoName)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


agnivade:
I have tested the `cloneRepo` function manually and verified that it works. But got a weird error while creating a cherry-pick end to end. https://github.com/mattermosttest/mattermost-server/pull/5#issuecomment-689601165. I am assuming this is some issue with my setup. Any ideas as to what's wrong? @ashishbhate

server/cherry_pick.go
```diff
@@ -204,20 +211,22 @@ func (s *Server) doCherryPick(ctx context.Context, version string, milestoneNumb
 	cmd.Env = append(
 		os.Environ(),
 		os.Getenv("PATH"),
-		fmt.Sprintf("DRY_RUN=%t", false),
```
agnivade, line 207: @isacikgoz - Adding this was incorrect. :P The `DRY_RUN` env var just checks whether it's empty or not. Setting it to false would actually enable dry run. 

It's good that we didn't deploy yet. :D

isacikgoz: We need to go over the script and simplify it 😅

isacikgoz (changes requested):
Looks good overall, spotted an issue and have a question below 👇

agnivade:
> I have tested the cloneRepo function manually and verified that it works. But got a weird error while creating a cherry-pick end to end. mattermosttest/mattermost-server#5 (comment). I am assuming this is some issue with my setup. Any ideas as to what's wrong? @ashishbhate

I tracked it down to an issue with `hub`. Not sure what's happening but I can't seem to create a PR. Searching the internets show either a login issue or that the branch is not present. Atleast it does not look to be related to my code as the git cloning and setting the upstream works correctly.

server/cherry_pick.go
```diff
@@ -316,15 +328,65 @@ func (s *Server) addAssignee(ctx context.Context, newPRNumber int, pr *model.Pul
...
+	return nil
+}
+
+func cloneRepo(ctx context.Context, dir, upstreamSlug, originSlug, repoName string) error {
+	// Set username and email first.
+	cmd := exec.CommandContext(ctx, "git", "config", "user.name")
+	if out, err := runCommandWithOutput(cmd, dir); err != nil {
+		return err
+	} else if out == "" { // this means username is not set
+		cmd = exec.CommandContext(ctx, "git", "config", "--global", "user.name", "mattermost-build")
```
khos2ow, line 345: These need to be checked and set per cloned repository, `--global` gitconfig is not accessible.
And because of this I think you need to move this block (`git config user.{name|email}`) after the repos is cloned and inside the cloned local folder.

khos2ow (changes requested):
I have one blocker comment below otherwise LGTM.

server/cherry_pick.go
agnivade: Ah my bad. Will change.

khos2ow: `cd <just-cloned-folder>` command seems to be missing now.

agnivade: It's not. `cmd.Dir = dir` is set and dir is `filepath.Join(dir, repoName)` ;)

khos2ow (approved):

server/cherry_pick.go
```diff
@@ -316,15 +328,65 @@ func (s *Server) addAssignee(ctx context.Context, newPRNumber int, pr *model.Pul
...
+	cmd := exec.CommandContext(ctx, "git", "checkout", "master")
+	if err := runCommand(cmd, dir); err != nil {
+		return err
+	}
+	return nil
+}
+
+func cloneRepo(ctx context.Context, dir, upstreamSlug, originSlug, repoName string) error {
+	// Clone repo
+	cmd := exec.CommandContext(ctx, "git", "clone", "--depth=1", "git@github.com:"+originSlug+".git")
```
ashishbhate, line 341: I think `--depth=1` also implies `--single-branch`, so isn't only the master branch pulled? Doesn't this cause a failure when we check out the release branch in the cherry-pick.sh script?

Even if `--single-branch` isn't implied (older versions of git), then we still wont be able to checkout branches that are deeper than the given revisions old

The cherry-pick.sh script does a `git remote update`, but that doesn't fetch other branches or older commits.  This means that the local checkout might not have the commit we want to cherry-pick, if that commit is sufficiently old. 

I'm not a git expert, so my comments above need to be double checked.

ashishbhate (commented):
@agnivade it's possible the issue you're seeing is because of some caching/slow processing at Github's end. Maybe Github hasn't detected the pushed branch yet, so it can't make PR from it.

server/cherry_pick.go
agnivade: Yeah, it means only master branch. But when I ran the script, it downloaded some more stuff by itself and was able to successfully push a branch, just failing in the hub command. I didn't check how it did that though. Nevertheless, I'll remove `--depth=1` to be safe.

agnivade:
> it's possible the issue you're seeing is because of some caching/slow processing at Github's end. Maybe Github hasn't detected the pushed branch yet, so it can't make PR from it.

I added a sleep 5 after the git push, but still the results were the same. :man_shrugging:

server/cherry_pick.go
```diff
@@ -316,15 +328,65 @@ func (s *Server) addAssignee(ctx context.Context, newPRNumber int, pr *model.Pul
...
+		if err = runCommand(cmd, filepath.Join(dir, repoName)); err != nil {
+			return err
+		}
+	}
+
+	cmd = exec.CommandContext(ctx, "git", "config", "user.email")
+	if out, err := runCommandWithOutput(cmd, filepath.Join(dir, repoName)); err != nil {
+		return err
+	} else if out == "" { // this means email is not set
+		cmd = exec.CommandContext(ctx, "git", "config", "user.email", "build@mattermost.com")
```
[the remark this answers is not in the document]
agnivade: I don't really see the email available in the config. It's there in Jenkins credentials, but I don't want to overload that. Do you see another way to get it from the config.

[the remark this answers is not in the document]
ashishbhate: Yep, sorry, looks like email isn't in the config. 

My vote would be to add it to the mattermod config similar to `GithubUsername` - something like `GithubEmail` or `GithubUserEmail`

ashishbhate (approved):
Looks great! 

Not sure why hub is failing though. I've never seen that error. It looks like an issue on Hub/Github's side. I think this PR is good to go.

isacikgoz (approved):
Looks good, we need to check if `defaultRequestTimeout*time.Second` will be enough for a single cherry pick request.

agnivade:
@isacikgoz - That's why I have multiplied it by 2 https://github.com/mattermost/mattermost-mattermod/pull/233/files#diff-118aa4c25aa26b3cc83311989dcc5448R44 :)

PR #234, opened 2020-09-10
MM-27539: Use pagination for s.waitForWorkflowID

agnivade:
Look for the page token and go through all the pages
until we get the workflow id.

We also bring back the API retry logic which got removed
when we moved to use the circleCI API.

https://mattermost.atlassian.net/browse/MM-27539

### DIFF
--- a/server/circleci.go
+++ b/server/circleci.go
@@ -203,15 +203,33 @@ func (s *Server) waitForWorkflowID(ctx context.Context, id string, workflowName 
 		case <-ctx.Done():
 			return "", errors.New("timed out trying to fetch workflow")
 		case <-ticker.C:
-			wfList, err := s.CircleCiClientV2.GetPipelineWorkflowWithContext(ctx, id, "")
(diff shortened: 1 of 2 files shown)


server/circleci.go
```diff
@@ -203,15 +203,33 @@ func (s *Server) waitForWorkflowID(ctx context.Context, id string, workflowName
...
+
+				if workflowID != "" {
+					return workflowID, nil
+				}
+
+				if wfList.NextPageToken != "" {
+					token = wfList.NextPageToken
+				} else {
 					break
 				}
```
isacikgoz, line 234: 😎 
```suggestion
				if wfList.NextPageToken == "" {
					break
				}
				token = wfList.NextPageToken
```

isacikgoz (approved):
LGTM 👍

metanerd (approved):
Looks good to me

ethervoid (approved):
LGTM!

PR #235, opened 2020-09-10
Bump page number while continuing in case of error

agnivade:
If a rate limit error occurs, we would just blindly retry that request
without incrementing the page number. This would keep retrying the same page
until GH would eventually return a proper response or return NextPage = 0.

A better way is to move on to the next page so that even if an error occurs,
we eventually converge to getting out of the loop.

While here, we also add some sleeps to better manage rate limit errors.
Ideally, we would just tune the burst rate, but for now we are sharing the same config
for both Job server and API server. To keep things simple, we change the code.


isacikgoz (approved):
Looks good to me 👍

agnivade:
Actually, thinking more about it, every single request is for a different page and is unlikely to be cached because new issues or new PRs would have been created. I think it's better to move the 200ms sleep in between requests rather than just when an error occurs.

ethervoid (approved):
LGTM

PR #236, opened 2020-09-10
Add version info, commit hash and build date to binary

khos2ow:
#### Summary

Add version information including actual version, commit hash, and build date to the built binary through ldfalgs on build time.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,11 +9,11 @@ BUILD_DATE   ?= $(shell date +%FT%T%z)
 CUR_VERSION  ?= $(shell git describe --tags --exact-match 2>/dev/null || git describe --tags 2>/dev/null || echo "v0.0.0-$(COMMIT_HASH)")
 
 # Go variables
-GOLDFLAGS   :="
-GOLDFLAGS   += -X github.com/$(AUTHOR)/$(NAME)/version.version=$(CUR_VERSION)
-GOLDFLAGS   += -X github.com/$(AUTHOR)/$(NAME)/version.commitHash=$(COMMIT_HASH)
-GOLDFLAGS   += -X github.com/$(AUTHOR)/$(NAME)/version.buildDate=$(BUILD_DATE)
-GOLDFLAGS   +="
+LDFLAGS :="
+LDFLAGS += -X github.com/$(AUTHOR)/$(NAME)/version.version=$(CUR_VERSION)
+LDFLAGS += -X github.com/$(AUTHOR)/$(NAME)/version.commitHash=$(COMMIT_HASH)
+LDFLAGS += -X github.com/$(AUTHOR)/$(NAME)/version.buildDate=$(BUILD_DATE)
+LDFLAGS +="
 
 GO ?= $(shell command -v go 2> /dev/null)
 
@@ -73,11 +73,11 @@ build: build-mattermod build-jobserver
 
 build-mattermod: clean
 	@echo Building mattermod
-	$(GO) build -ldflags $(GOLDFLAGS) -o dist/mattermod ./cmd/mattermost-mattermod
+	$(GO) build -ldflags $(LDFLAGS) -o dist/mattermod ./cmd/mattermost-mattermod
 
 build-jobserver: clean
 	@echo Building mattermod
-	$(GO) build -ldflags $(GOLDFLAGS) -o dist/jobserver ./cmd/jobserver
+	$(GO) build -ldflags $(LDFLAGS) -o dist/jobserver ./cmd/jobserver
 
 # Docker variables
 DEFAULT_TAG  ?= $(shell git describe --tags --exact-match 2>/dev/null || git rev-parse --short HEAD 2>/dev/null)
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,22 @@
-GO ?= $(shell command -v go 2> /dev/null)
+# Project variables
+NAME        := mattermost-mattermod
+AUTHOR      := mattermost
+URL         := https://github.com/$(AUTHOR)/$(NAME)
+
+# Build variables
+COMMIT_HASH  ?= $(shell git rev-parse --short HEAD 2>/dev/null)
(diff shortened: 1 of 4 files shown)


version/version.go
```diff
@@ -0,0 +1,43 @@
+package version
```
agnivade, line 1: It's better if we perform this task of setting ldflag variables from main, and then set them in the config and pass them to the server struct. And then we can return them in the ping response.

Creating a separate package just to have a single function and some variables is probably an overkill.

hack/release.sh
```diff
@@ -76,3 +76,15 @@ git tag --annotate --message "v${RELEASE_VERSION} Release" "v${RELEASE_VERSION}"
...
 git push origin v${RELEASE_VERSION}
+
+# Bump the next version in version.go
+NEXT_VERSION=$(echo "${RELEASE_VERSION}" | sed 's/^v//' | awk -F'[ .]' '{print $1"."$2+1".0"}')
+sed -i -E 's|v'${RELEASE_VERSION}'|v'${NEXT_VERSION}'-alpha|g' version/version.go
+
+# Commit changes
+printf "\033[36m==> %s\033[0m\n" "Bump version to v${NEXT_VERSION}-alpha"
+git add version/version.go
+git commit -m "Bump version to v${NEXT_VERSION}-alpha"
```
agnivade, line 87: It's a bit odd to see versions being hardcoded inside the code. I don't see this being done on any other services. IMO, code should be concerned with just the task it's doing and not contain a hardcoded version string in it. That can be done via git and then injected at runtime (exactly like we are doing via ldflags).

So my suggestion would be to drop the modification of the version file and just dynamically inject the version during building.

agnivade (commented):
This is a great start and was sorely needed. I just have some small concerns with the code layout and the implementation. Thanks!

Makefile
```diff
@@ -1,4 +1,22 @@
...
+AUTHOR      := mattermost
+URL         := https://github.com/$(AUTHOR)/$(NAME)
+
+# Build variables
+COMMIT_HASH  ?= $(shell git rev-parse --short HEAD 2>/dev/null)
+BUILD_DATE   ?= $(shell date +%FT%T%z)
+CUR_VERSION  ?= $(shell git describe --tags --exact-match 2>/dev/null || git describe --tags 2>/dev/null || echo "v0.0.0-$(COMMIT_HASH)")
+
+# Go variables
+GOLDFLAGS   :="
```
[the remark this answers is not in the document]
khos2ow: I specifically didn't name it `LDFLAGS` to prevent environment variable collision that one might already set in their terminal and let `GOLDFLAGS` be dedicated to this Makefile targets. I'm completely fine with changing it though.

Makefile
```diff
@@ -1,4 +1,22 @@
...
+GOLDFLAGS   += -X github.com/$(AUTHOR)/$(NAME)/version.version=$(CUR_VERSION)
+GOLDFLAGS   += -X github.com/$(AUTHOR)/$(NAME)/version.commitHash=$(COMMIT_HASH)
+GOLDFLAGS   += -X github.com/$(AUTHOR)/$(NAME)/version.buildDate=$(BUILD_DATE)
+GOLDFLAGS   +="
+
+GO       ?= $(shell command -v go 2> /dev/null)
+GOBUILD  ?= $(GO) build -ldflags $(GOLDFLAGS)
```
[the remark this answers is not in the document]
khos2ow: That's a completely fair point, I extract it up top to basically DRY. Will change it as you suggested.

hack/release.sh
khos2ow: I always put a hardcoded dev version in my projects, mainly to give visibility of the _version_ if someone `go get` rather than download the binary (released or rc or ...)
I can see the point here, because most probably no one will ever _go get_ this. But on the other hand that variable never gets changed manually and the release script changes it and also the injection upon release will override that anyways. So I'm kinda 2-ish/5 to keep it.

version/version.go
khos2ow: That's the exact reason I did put them in their own package. Adding version metadata to Config to pass it around does not make much sense to me. These information are not _configuration_, rather metadata of the binary.

Also I don't quite like adding random top level variables in main just for the sake of ldflags to work.

I'm 4/5 to keep it like this but open to change it.

Makefile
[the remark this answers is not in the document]
agnivade: My Makefile knowledge is very superficial, but I think this is just a Makefile variable and should not collide if an environment variable with the same name is set.

version/version.go
agnivade: Ok.

hack/release.sh
agnivade: The visibility of the version is not the responsibility of the code IMO. I'd be interested to see if this is documented as a best practice somewhere.

If it's a library, the version is tracked in the go.mod file, otherwise if it's a main binary, it is tracked in git. Exactly how it's done in all of our projects.

I think just embedding the version from git should be good enough.

metanerd (approved):
LGTM

agnivade (approved):
Nothing blocking. Thanks for this :)

version/version.go
isacikgoz: Since we already have other assets we can include another one like `version.json` and pass all these ldflags into it at `make build` stage. Then we can just read it whenever we hit the ping handler. In fact, we can simply read it binary and attach to response as `json.RawMessage`. With this we won't need to manage `version.Info` struct at all.

isacikgoz (approved):
Looks good. I have another proposal to the discussion above. If it makes sense to you, you can consider it 😄 Anyway this is also good to go.

version/version.go
khos2ow: Since version information is more of metadata than asset I think it'd be better to keep it in `version.go` file and let the ldflags  - on build time - inject the values to it.

PR #237, opened 2020-09-15
MM-26794: Implement a DB migration plan

agnivade:
This PR removes all the gorp related code which would perform DB changes
automatically from models, and implements a raw SQL file based approach to migation.

We use the golang-migrate library to implement migrations. And as the first migration,
we create the base schema of the DB itself.

More info is added in the README.

A CLI tool is also added to help with rolling back schema versions.

After this, we can slowly remove the gorp code from the codebase. Now gorp does not serve
any other purpose than just an ORM.

https://mattermost.atlassian.net/browse/MM-26794


agnivade:
@mgdelacroix / @streamer45 - cc'd you guys because this is an important PR which introduces DB migrations as a first step to get us away from gorp. We plan to use mattermod as a test bed to see how to successfully move away from gorp, and then apply the learnings to the main server repo.

Please take a look and let me know if you have any questions.

ethervoid (commented):
Some comments

store/migrations/000001_base.down.sql
```diff
@@ -0,0 +1,7 @@
+BEGIN;
+
+DROP TABLE IF EXISTS `Issues`;
+DROP TABLE IF EXISTS `PullRequests`;
+DROP TABLE IF EXISTS `Spinmint`;
+
+COMMIT;
```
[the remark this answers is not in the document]
agnivade: Doesn't matter much, but okay. :)

store/sql_store.go
```diff
@@ -325,7 +109,39 @@ func (ss *SQLStore) Spinmint() SpinmintStore {
...
+	if err != nil {
+		mlog.Critical("Failed to create source instance", mlog.Err(err))
+		os.Exit(1)
+	}
+
+	m, err := migrate.NewWithInstance("go-bindata", srcDriver, "mysql", dbDriver)
+	if err != nil {
+		mlog.Critical("Failed to create db instance", mlog.Err(err))
+		os.Exit(1)
+	}
```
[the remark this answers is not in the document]
agnivade: That will close the db. So the application will fail to run. Not something we want :joy:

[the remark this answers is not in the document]
ethervoid: Hahaha ok. I thought it was part of the migration code

ethervoid (approved):
LGTM! Eager to see it working

README.md
```diff
@@ -109,3 +109,69 @@ docker run --name mattermod-mysql -p 3306:3306 -e MYSQL_ROOT_PASSWORD=mattermod
...
+
+And invert the condition if you want to modify if the column exists.
+
+Other solutions include using a stored procedure, but the mysql driver doesn't support it well.
+
+### Rolling back schema versions
+
+If we need to roll back mattermod to an earlier version, then we also need to check if there have been any schema changes in between. If there is, then a CLI tool is provided that can be used to manually run forward or reverse migrations as necessary.
+
+For example, if we have migrated to version 3 (available from `select version from schema_migrations order by version desc limit 1;`), and want to roll back to 2, then we can enter:
```
streamer45, line 171:
```suggestion
For example, if we have migrated to version 3 (available from `SELECT version FROM schema_migrations ORDER BY version DESC LIMIT 1;`), and want to roll back to 2, then we can enter:
```

store/sql_store.go
```diff
@@ -325,7 +109,39 @@ func (ss *SQLStore) Spinmint() SpinmintStore {
...
+	// Run migration
+	err = m.Up()
+	// We ignore if there is no change and if file does not exist.
+	// The latter occurs if we have rolled back to older code without running down migrations.
+	// This is not ideal, but not critical enough to bail execution.
+	if err != nil && err != migrate.ErrNoChange && err != os.ErrNotExist {
+		mlog.Critical("Failed to migrate DB", mlog.Err(err))
+		os.Exit(1)
```
streamer45, line 144: I can't believe I am saying this but wouldn't it be more *idiomatic* to use `errors.Is()`?

cmd/mattermost-mattermod/migrations.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	m, err := migrate.NewWithInstance("go-bindata", srcDriver, "mysql", dbDriver)
+	if err != nil {
+		return fmt.Errorf("failed to create db instance: %v", err)
+	}
+	defer m.Close()
+
+	err = m.Migrate(uint(migrateVersion))
+	if err != nil && err != migrate.ErrNoChange {
```
streamer45, line 48: Same here.

cmd/mattermost-mattermod/migrations.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"github.com/mattermost/mattermost-mattermod/store/migrations"
+
+	_ "github.com/go-sql-driver/mysql" // Load MySQL Driver
+	"github.com/golang-migrate/migrate/v4"
+	"github.com/golang-migrate/migrate/v4/database/mysql"
+	bindata "github.com/golang-migrate/migrate/v4/source/go_bindata"
+)
+
+func runMigrations(driverName, dataSource string, migrateVersion int) error {
```
streamer45, line 18: I see that most of this is duplicated in the store with the main difference being that there we go up only.
Would it make sense to have a more generic version over there which can be also re-used here?

streamer45 (commented):
Looks very good :+1: 
Left a few minor suggestions/comments.

README.md
```diff
@@ -109,3 +109,69 @@ docker run --name mattermod-mysql -p 3306:3306 -e MYSQL_ROOT_PASSWORD=mattermod
 6. For cherry-picking to work, [hub](https://github.com/github/hub) needs to be installed in the system, and the script from `hacks/cherry-pick.sh` should be placed at `/app/scripts/cherry-pick.sh`.
 
 7. Start up Mattermod server.
+
+## Migrations
+
+We use the [golang-migrate](https://github.com/golang-migrate/migrate) library combined with code generation to handle migrations. To add a new migration in your code:
+
+1. **Create files:** Go to the `store/migrations` directory and create 2 files by incrementing the latest sequence number of the files in that directory. For example, if the last number is 000004, then please 2 files with names `000005_{migration_name}.up.sql` and `000005_{migration_name}.down.sql`.
```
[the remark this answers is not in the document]
agnivade: I'm always kind :smile:  It's missing a verb.

store/sql_store.go
agnivade: Haha .. yeah for `ErrNotExist` definitely. :+1: May not be necessary for `ErrNoChange` since it's directly returned from the library itself and is publicly documented.

cmd/mattermost-mattermod/migrations.go
agnivade: It's just creating the database and source driver which can be de-duped. And we'll have to do the error checking anyways. So it'll just replace 2 function calls with 1. Not sure if it's worth it.

streamer45: I think all the `m *Migrate` initialization part could be a separate function to be used in the two places. One will run `m.Migrate()` and the other `m.Up()`. But yes, not a big deal. It's also fine as it is.

README.md
```diff
@@ -109,3 +109,69 @@ docker run --name mattermod-mysql -p 3306:3306 -e MYSQL_ROOT_PASSWORD=mattermod
...
+migrate -database "mysql://<user>:<pass>@tcp(<host>:3306)/<db>?parseTime=true&multiStatements=true" -path store/migrations/ force 1
+migrate -database "mysql://<user>:<pass>@tcp(<host>:3306)/<db>?parseTime=true&multiStatements=true" -path store/migrations/ up 1
+migrate -database "mysql://<user>:<pass>@tcp(<host>:3306)/<db>?parseTime=true&multiStatements=true" -path store/migrations/ down 1
+migrate -database "mysql://<user>:<pass>@tcp(<host>:3306)/<db>?parseTime=true&multiStatements=true" -path store/migrations/ force 2
+migrate -database "mysql://<user>:<pass>@tcp(<host>:3306)/<db>?parseTime=true&multiStatements=true" -path store/migrations/ down 1
+migrate -database "mysql://<user>:<pass>@tcp(<host>:3306)/<db>?parseTime=true&multiStatements=true" -path store/migrations/ up 1
+```
+
+**Special note on existing columns**: MySQL does not have a native way to check if a column exists or not (yay MySQL). So if a migration adds a new column, it has to use the `information_schema.columns` table to check whether a column exists or not. You can use the following hack to do such a thing:
```
isacikgoz, line 141: 🤣

bin/.gitignore
```diff
@@ -0,0 +1,3 @@
+# Folder required for build process
```
isacikgoz, line 1:
```suggestion
# Directory required for build process
```

cmd/mattermost-mattermod/main.go
```diff
@@ -31,6 +39,15 @@ func main() {
 		os.Exit(1)
 	}
 
+	if migrateVersion != -1 {
+		err = runMigrations(config.DriverName, config.DataSource, migrateVersion)
```
isacikgoz, line 43: Nit: we could have separate main for this under something like `/cmd/mattermod-migrator`.

isacikgoz (approved):
Thanks for doing such an overhaul @agnivade great work! I have some minor comments.

cmd/mattermost-mattermod/main.go
agnivade: Fair enough. I'll make the changes. :+1:

agnivade:
@khos2ow - We are releasing a new binary with this PR. How do you prefer to deploy this? Should I add a create a new Dockerfile for this? Or you prefer to just build the binary and run the commands?

streamer45 (approved):

khos2ow:
@agnivade when and how this new binary should run? Is this only for the operation side to roll back?

agnivade:
>  Is this only for the operation side to roll back?

Yep. Only when you want to roll back to an earlier version where you would want to revert schema changes, you'll need to run the binary as:

`migrator -config config/config-mattermod.json -migration_version <target_version>`

Can be run from anywhere as long as it can access the prod DB.

khos2ow:
So in that case I don't think we need a containerized version of it rather a binary (for different os and arch) for operation folks to be able to download from GH release page.

agnivade:
Will do, thanks!

mgdelacroix (approved):
This is awesome!! :rocket: Added a couple of nits, but no blocker

cmd/migrator/main.go
```diff
@@ -0,0 +1,44 @@
...
+		os.Exit(1)
+	}
+
+	if migrateVersion != -1 {
+		err = runMigrations(config.DriverName, config.DataSource, migrateVersion)
+		if err != nil {
+			mlog.Error("Failed to run migrations", mlog.Err(err))
+			os.Exit(1)
+		}
+	}
```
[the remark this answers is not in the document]
agnivade: Good point.

agnivade:
Waiting until 0.4.1 is deployed and stable.

PR #238, opened 2020-09-15
MM-27487: Use PR milestone info from DB

agnivade:
We store the milestone info in the DB and then use that to perform
the cherry-picking action. This saves a call to GitHub.

To achieve this, we need to differentiate between an issue and a PR
in the issue event handler because GitHub treats all PRs to be issues.

If a PR milestone is changed, we just update the DB with the latest PR info.

https://mattermost.atlassian.net/browse/MM-27487


agnivade:
This needs to go after the migration PR. But it's good to review.

server/pr_from_issue_handler.go
```diff
@@ -0,0 +1,41 @@
...
+	// We update the milestone that we have from the issue event and merge it with the PR.
+	// This is necessary to work around caching issues with GitHub.
+	prGitHub.Milestone = event.Issue.GetMilestone()
+
+	_, err = s.GetPullRequestFromGithub(ctx, prGitHub)
+	if err != nil {
+		mlog.Error("Error in saving the PR", mlog.Err(err))
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
```
ashishbhate, line 40: Is the query to Github to get the PR always necessary?
If we already have the PR saved in the DB how about just loading it from the DB and updating the milestone?

ashishbhate (commented):
LGTM! Great job navigating Github's issue/PR conflation.

We can drop fetching the PR again in the `Server.checkIfNeedCherryPick` method (and possible other places), is that change intended for a separate PR?

agnivade:
> We can drop fetching the PR again in the Server.checkIfNeedCherryPick method (and possible other places), is that change intended for a separate PR?

Oh lol, completely forgot to remove that!

server/pr_from_issue_handler.go
agnivade: Hmm .. that's a good point. Will update.

agnivade:
@ashishbhate PTAL

server/pr_from_issue_handler.go
```diff
@@ -0,0 +1,34 @@
...
+
+func (s *Server) prFromIssueHandler(event *issueEvent, w http.ResponseWriter) {
+	oldPR, err := s.Store.PullRequest().Get(event.Repo.GetOwner().GetLogin(),
+		event.Repo.GetName(),
+		event.Issue.GetNumber())
+	if err != nil {
+		mlog.Error("Error in getting PR from DB", mlog.Err(err))
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
```
ashishbhate, line 21: Non-blocker: 
I think we should get the PR from github if we don't already have it, instead of returning error directly.

server/pr_from_issue_handler_test.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/google/go-github/v32/github"
+	"github.com/mattermost/mattermost-mattermod/model"
+	stmock "github.com/mattermost/mattermost-mattermod/store/mocks"
+	"github.com/stretchr/testify/require"
+)
+
+func TestPRFromIssueHandler(t *testing.T) {
+	ctrl := gomock.NewController(t)
+
+	// ctxInterface := reflect.TypeOf((*context.Context)(nil)).Elem()
```
ashishbhate, line 24: :thinking:

ashishbhate (approved):

server/pr_from_issue_handler.go
agnivade: This handler will only be called when adding a milestone for an existing PR. A new PR would already be created in the DB from other handlers. So I think it should be fine. I will obviously monitor the logs if this happens and will add the case if needed.

server/pr_from_issue_handler_test.go
agnivade: :facepalm:

agnivade:
Gentle bump @isacikgoz

server/pr_from_issue_handler.go
```diff
@@ -0,0 +1,34 @@
...
+
+	// We update the milestone that we have from the issue event and merge it with the PR.
+	// This is necessary to work around caching issues with GitHub.
+	oldPR.MilestoneNumber = sql.NullInt64{Int64: int64(event.Issue.GetMilestone().GetNumber()), Valid: true}
```
isacikgoz, line 25: I got a panic here when I created a PR. (I added it to a milestone before creating it)

isacikgoz (changes requested):
Nicely done! Sorry for the delay on this. I tested it and caught a case below 👇 otherwise it is good to go!

server/pr_from_issue_handler.go
agnivade: Ohh .. what a bother. Our code returns nil error _and_ a nil PR if the row is not found .. :angry:

isacikgoz (approved):
Awesome 🎉

PR #240, opened 2020-09-22
Remove retrigger of circleci if needed

metanerd:
#### Summary
Removing triggering circleci if needed.


jupenur (approved):

agnivade (commented):
A question:

How are staff supposed to trigger CI from now on? Is there a new label like `EETests`?

PR #244, opened 2020-09-24
Simplify release publishing

streamer45:
#### Summary

PR adds a `publish-release` target to the `Makefile` that uses [Goreleaser](https://goreleaser.com/) to simplify the publishing of new releases.
This was used to generate [v0.5.0](https://github.com/mattermost/mattermost-mattermod/releases/tag/v0.5.0).

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,0 +1,70 @@
+before:
+  hooks:
+    - go mod download
+    - go mod verify
+builds:
+  - main: ./cmd/mattermost-mattermod
+    id: "mattermod"
+    binary: bin/mattermod
+    goos:
+      - linux
+    goarch:
+      - amd64
+    flags:
+      - -mod=readonly
+      - -trimpath
+    ldflags:
+      - -X github.com/mattermost/mattermost-mattermod/version.version=v{{.Version}}
+      - -X github.com/mattermost/mattermost-mattermod/version.commitHash={{.FullCommit}}
+      - -X github.com/mattermost/mattermost-mattermod/version.buildDate={{.Date}}
+    env:
+      - CGO_ENABLED=0
+  - main: ./cmd/jobserver
+    id: "jobserver"
+    binary: bin/jobserver
+    goos:
+      - linux
+    goarch:
+      - amd64
+    flags:
+      - -mod=readonly
+      - -trimpath
+    ldflags:
+      - -X github.com/mattermost/mattermost-mattermod/version.version=v{{.Version}}
+      - -X github.com/mattermost/mattermost-mattermod/version.commitHash={{.FullCommit}}
+      - -X github.com/mattermost/mattermost-mattermod/version.buildDate={{.Date}}
+    env:
+      - CGO_ENABLED=0
+  - main: ./cmd/migrator
+    id: "migrator"
+    binary: bin/migrator
+    goos:
+      - linux
+    goarch:
+      - amd64
+    flags:
+      - -mod=readonly
+      - -trimpath
+    ldflags:
+      - -X github.com/mattermost/mattermost-mattermod/version.version=v{{.Version}}
+      - -X github.com/mattermost/mattermost-mattermod/version.commitHash={{.FullCommit}}
+      - -X github.com/mattermost/mattermost-mattermod/version.buildDate={{.Date}}
+    env:
+      - CGO_ENABLED=0
+archives:
+  -
+    name_template: "{{ .ProjectName }}-v{{ .Version }}-{{ .Os }}-{{ .Arch }}"
+    wrap_in_directory: true
+    format: tar.gz
+    files:
+      - LICENSE
+      - README.md
+      - config/config-mattermod.default.json
+changelog:
+  sort: desc
+release:
+  github:
+    owner: mattermost
+    name: mattermost-mattermod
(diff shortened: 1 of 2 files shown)


agnivade (commented):
Can we also add the publish-release target after the patch, minor and major targets so that a release gets generated automatically once a tag is created? Or does it need the tag to be pushed first to GitHub? In which case, I'd prefer some `@echo` commands nudging the operator to run publish-release.

streamer45:
> Can we also add the publish-release target after the patch, minor and major targets so that a release gets generated automatically once a tag is created? Or does it need the tag to be pushed first to GitHub? In which case, I'd prefer some `@echo` commands nudging the operator to run publish-release.

The tag just has to exist but I see those targets do commit and push it so running `publish-release` after should work just fine.

agnivade (approved):

Makefile
```diff
@@ -164,6 +164,21 @@ minor: release
...
+		echo "goreleaser is not installed, do you want to download it? [y/N] " && read ans && [ $${ans:-N} = y ]; \
+		if [ $$ans = y ] || [ $$ans = Y ]  ; then \
+			curl -sfL https://install.goreleaser.com/github.com/goreleaser/goreleaser.sh | sh; \
+		else \
+			echo "aborting make publish-release."; \
+			exit -1; \
+		fi; \
+	else \
+		goreleaser --rm-dist; \
+	fi; \
```
isacikgoz, line 180: We need to remove this `else` block, If I don't have `goreleaser` and install it here, I need to run `make publish-release` again to make this work.

streamer45: Oh I see. It was meant to install and run at the same time, I'll move it after the `curl | sh` bit then.

streamer45: One a second though. Might as well keep it at the end since we exit on failure.

isacikgoz: Yeah, definitely :)

isacikgoz (approved):
Awesome! 🚀

.goreleaser.yml
```diff
@@ -0,0 +1,70 @@
+before:
+  hooks:
+    - go mod download
+    - go mod verify
+builds:
+  - main: ./cmd/mattermost-mattermod
```
khos2ow, line 6: We don't necessarily need to release this binary since we already publish docker image.

.goreleaser.yml
```diff
@@ -0,0 +1,70 @@
...
+    flags:
+      - -mod=readonly
+      - -trimpath
+    ldflags:
+      - -X github.com/mattermost/mattermost-mattermod/version.version=v{{.Version}}
+      - -X github.com/mattermost/mattermost-mattermod/version.commitHash={{.FullCommit}}
+      - -X github.com/mattermost/mattermost-mattermod/version.buildDate={{.Date}}
+    env:
+      - CGO_ENABLED=0
+  - main: ./cmd/jobserver
```
khos2ow, line 22: The same here.

.goreleaser.yml
```diff
@@ -0,0 +1,70 @@
...
+      - -X github.com/mattermost/mattermost-mattermod/version.version=v{{.Version}}
+      - -X github.com/mattermost/mattermost-mattermod/version.commitHash={{.FullCommit}}
+      - -X github.com/mattermost/mattermost-mattermod/version.buildDate={{.Date}}
+    env:
+      - CGO_ENABLED=0
+  - main: ./cmd/migrator
+    id: "migrator"
+    binary: bin/migrator
+    goos:
+      - linux
```
khos2ow, line 42: This is essentially what's needed and it's a tool for operators so we need to publish all available os (`darwin`, `windows`, `linux` and maybe just throwing `freebsd` out there too.)

.goreleaser.yml
```diff
@@ -0,0 +1,70 @@
...
+    goarch:
+      - amd64
```
khos2ow, line 44: It's nice to have `arm` too.

.goreleaser.yml
```diff
@@ -0,0 +1,70 @@
...
+      - CGO_ENABLED=0
+archives:
+  -
+    name_template: "{{ .ProjectName }}-v{{ .Version }}-{{ .Os }}-{{ .Arch }}"
+    wrap_in_directory: true
+    format: tar.gz
+    files:
+      - LICENSE
+      - README.md
+      - config/config-mattermod.default.json
```
khos2ow, line 62: We shouldn't bundle config with the release artifact, it gives an impression that that config is needed (which is not, it's just an in-tree sample)

Makefile
```diff
@@ -152,17 +152,31 @@ release:
...
+	@if ! [ -x "$$(command -v goreleaser)" ]; then \
+		echo "goreleaser is not installed, do you want to download it? [y/N] " && read ans && [ $${ans:-N} = y ]; \
+		if [ $$ans = y ] || [ $$ans = Y ]  ; then \
+			curl -sfL https://install.goreleaser.com/github.com/goreleaser/goreleaser.sh | sh; \
+		else \
+			echo "aborting make publish-release."; \
+			exit -1; \
+		fi; \
+	fi; \
+	goreleaser --rm-dist; \
```
khos2ow, line 179: Leftover `\` at the end of line.

khos2ow (commented):
It's already merged but I have some comments below.

PR #246, opened 2020-09-26
[MM-29076] Send welcome message for community PRs

hanzei:
#### Summary
For all community PRs we want to send a welcome message to educate them about the review process.

The actual message send to the user can be found at https://github.com/mattermost/platform-private/pull/284.

Are there repository we want to don't post such message for? e.g. https://github.com/mattermost/docs

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29076

### DIFF
--- a/server/cla.go
+++ b/server/cla.go
@@ -100,15 +100,6 @@ func isNameInCLAList(usersWhoSignedCLA []string, authorToTrim string) bool {
 	return false
 }
 
-func findNeedsToSignCLAComment(comments []*github.IssueComment, username string) (id int64, found bool) {
-	for _, comment := range comments {
-		if *comment.User.Login == username && strings.Contains(*comment.Body, "Please help complete the Mattermost") {
-			return *comment.ID, true
-		}
-	}
-	return 0, false
-}
-
 func (s *Server) createCLAPendingStatus(ctx context.Context, pr *model.PullRequest) {
 	status := &github.RepoStatus{
 		State:       github.String(statePending),
--- a/server/cla.go
+++ b/server/cla.go
@@ -16,6 +16,8 @@ import (
 	"github.com/mattermost/mattermost-server/v5/mlog"
 )
 
+// handleCheckCLA checks if the author of a pull request has signed the CLA and sets a status accordingly.
+// Returns true, if the user hasn't signed yet.
 func (s *Server) handleCheckCLA(ctx context.Context, pr *model.PullRequest) (bool, error) {
 	if pr.State == model.StateClosed {
 		return false, nil
--- a/server/cla.go
+++ b/server/cla.go
@@ -10,16 +10,15 @@ import (
 	"net/http"
 	"strconv"
 	"strings"
-	"time"
 
 	"github.com/google/go-github/v32/github"
 	"github.com/mattermost/mattermost-mattermod/model"
 	"github.com/mattermost/mattermost-server/v5/mlog"
 )
 
-func (s *Server) handleCheckCLA(ctx context.Context, pr *model.PullRequest) error {
+func (s *Server) handleCheckCLA(ctx context.Context, pr *model.PullRequest) (bool, error) {
 	if pr.State == model.StateClosed {
-		return nil
+		return false, nil
 	}
 
 	s.createCLAPendingStatus(ctx, pr)
@@ -41,29 +40,15 @@ func (s *Server) handleCheckCLA(ctx context.Context, pr *model.PullRequest) erro
 			Context:     github.String(s.Config.CLAGithubStatusContext),
 		}
 		mlog.Debug("will succeed CLA status for excluded user", mlog.String("user", username))
-		return s.createRepoStatus(ctx, pr, status)
+		return false, s.createRepoStatus(ctx, pr, status)
 	}
 
 	body, err := s.getCSV(ctx)
 	if err != nil {
-		return err
+		return false, nil
 	}
 
 	if !isNameInCLAList(strings.Split(string(body), "\n"), username) {
-		comments, err := s.getComments(ctx, pr.RepoOwner, pr.RepoName, pr.Number)
-		if err != nil {
-			return fmt.Errorf("failed fetching comments: %w", err)
-		}
-		_, found := findNeedsToSignCLAComment(comments, s.Config.Username)
-		if !found {
-			go func() {
-				ctx2, cancel := context.WithTimeout(context.Background(), defaultRequestTimeout*time.Second)
-				defer cancel()
-				if err = s.sendGitHubComment(ctx2, pr.RepoOwner, pr.RepoName, pr.Number, strings.Replace(s.Config.NeedsToSignCLAMessage, "USERNAME", "@"+username, 1)); err != nil {
-					mlog.Warn("Error while commenting", mlog.Err(err))
-				}
-			}()
-		}
 		status := &github.RepoStatus{
 			State:       github.String(stateError),
 			Description: github.String(fmt.Sprintf("%v needs to sign the CLA", username)),
@@ -71,7 +56,7 @@ func (s *Server) handleCheckCLA(ctx context.Context, pr *model.PullRequest) erro
 			Context:     github.String(s.Config.CLAGithubStatusContext),
 		}
 		mlog.Debug("will post error on CLA", mlog.String("user", username))
-		return s.createRepoStatus(ctx, pr, status)
+		return true, s.createRepoStatus(ctx, pr, status)
 	}
 
 	status := &github.RepoStatus{
@@ -81,7 +66,7 @@ func (s *Server) handleCheckCLA(ctx context.Context, pr *model.PullRequest) erro
 		Context:     github.String(s.Config.CLAGithubStatusContext),
 	}
 	mlog.Debug("will post success on CLA", mlog.String("user", username))
-	return s.createRepoStatus(ctx, pr, status)
+	return false, s.createRepoStatus(ctx, pr, status)
 }
 
 func (s *Server) getCSV(ctx context.Context) ([]byte, error) {
--- a/server/community.go
+++ b/server/community.go
@@ -4,12 +4,14 @@
 package server
 
 import (
+	"bytes"
 	"context"
-	"strings"
+	"text/template"
 	"time"
 
 	"github.com/mattermost/mattermost-mattermod/model"
 	"github.com/mattermost/mattermost-server/v5/mlog"
+	"github.com/pkg/errors"
 )
 
 func (s *Server) addHacktoberfestLabel(ctx context.Context, pr *model.PullRequest) {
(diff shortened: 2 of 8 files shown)


hanzei:
@cpanato I've include you as a reviewer because you brought this idea up originally.

jasonblais (approved):
Thanks! I assume this message is also posted when the contributor hasn't signed a CLA?

If yes, I will adjust it slightly in https://github.com/mattermost/platform-private/pull/284, so that the two messages (welcome + CLA), won't have duplicate information.

jasonblais:
@hanzei Actually, I forgot to comment on this:

> Are there repository we want to don't post such message for?

My first inclination is to post this to mattermost-server, mattermost-mobile, mattermost-webapp, desktop, mattermost-redux, mattermost-plugin-* repositories.

For most others, the code review process we're referencing does not apply.

server/pull_request_handler.go
```diff
@@ -60,7 +60,8 @@ func (s *Server) pullRequestEventHandler(w http.ResponseWriter, r *http.Request)
 			mlog.Debug("Triggered CircleCI", mlog.String("repo", pr.RepoName), mlog.Int("pr", pr.Number), mlog.String("fullname", pr.FullName))
 		}
 
-		s.addHacktoberfestLabel(ctx, pr)
+		go s.postPRWelcomeMessage(ctx, pr)
+		go s.addHacktoberfestLabel(ctx, pr)
```
agnivade, line 64: Any reason these are spun in goroutines? Unless they are long-running tasks, it would be better to just run them sequentially.

server/community.go
```diff
@@ -32,3 +33,18 @@ func (s *Server) addHacktoberfestLabel(ctx context.Context, pr *model.PullReques
 		return
 	}
 }
+
+func (s *Server) postPRWelcomeMessage(ctx context.Context, pr *model.PullRequest) {
```
agnivade, line 37: Please add a test case for this.

server/community.go
```diff
@@ -32,3 +33,18 @@ func (s *Server) addHacktoberfestLabel(ctx context.Context, pr *model.PullReques
...
+		return
+	}
+
+	msg := strings.ReplaceAll(s.Config.PRWelcomeMessage, "USERNAME", "@"+pr.Username)
+
+	err := s.sendGitHubComment(ctx, pr.RepoOwner, pr.RepoName, pr.Number, msg)
+	if err != nil {
+		mlog.Warn("Error while commenting PR welcome message", mlog.Err(err))
+		return
+	}
```
agnivade, line 49: Please propagate this error upwards and handle it at the call site.

server/community.go
```diff
@@ -32,3 +33,18 @@ func (s *Server) addHacktoberfestLabel(ctx context.Context, pr *model.PullReques
...
 	}
 }
+
+func (s *Server) postPRWelcomeMessage(ctx context.Context, pr *model.PullRequest) {
+	// Only post welcome Message for community member
+	if s.IsOrgMember(pr.Username) {
+		return
+	}
+
+	msg := strings.ReplaceAll(s.Config.PRWelcomeMessage, "USERNAME", "@"+pr.Username)
```
agnivade, line 43: I know we use "USERNAME" as a convention and just replace it. But it will be better if we can use `text/template`s for this. Would you be open to making that change here: https://github.com/mattermost/platform-private/pull/284? We can open more HW tickets changing the rest of the occurences to use `text/template`s

agnivade (changes requested):
Thanks. Please take a look at the comments.

A general thought: this will comment on all subsequent PRs for community members even after their first PR. I am 0/5 on the matter, but it might be a bit repetitive if every time someone opens a PR, they see this message. It's possible to store a flag for this in the DB for the PR, and next time, check the DB if the creator of the PR has a row where this flag is set, we can skip it. We don't need to implement it though. Just saying that _if_ we want something like this, it's possible.

server/pull_request_handler.go
hanzei: No they are not long running tasks. Are there downsides of using a go routine?

server/community.go
hanzei: Done

server/community.go
hanzei: Done

server/community.go
hanzei: While I see that using `text/template` is the "right" thing to do, I'm leaning toward consistency here. 1/5 that I would like to changes all occurrences of  `USERNAME` at the same time.

agnivade: Ok.

server/pull_request_handler.go
agnivade: Randomly spinning up goroutines is never a good idea. Whenever we spin up a goroutine, we need to think of how will it be shut down, what's the cancelling mechanism, so on and so forth.

An immediate bug here if we would create goroutines is the context. Since the same context is passed to a goroutine, it will get cancelled the moment it exits the handler. And therefore, it will cause all actions performed in the goroutine to be cancelled. This would need new contexts to be set, which loses out on the original point of being able to set the correct timeouts.

So more things go out of control as we randomly spin up goroutines. The proper way is to have a job queue to handle the entire request asynchronously. But that's a long, long way ahead.

server/community.go
```diff
@@ -32,3 +33,19 @@ func (s *Server) addHacktoberfestLabel(ctx context.Context, pr *model.PullReques
 		return
 	}
 }
+
+func (s *Server) PostPRWelcomeMessage(ctx context.Context, pr *model.PullRequest) error {
```
agnivade, line 37: Any reason this is exported now? The test file is in the same package, so it will still be able to access if it's not exported.

server/community.go
isacikgoz: I'm OK using this with `strings.Replace()` for the consistency. However, as a nice improvement, I'd like to see all of the `xMessage` in the `config.json` to be moved into something like `messages.go` or in a text file as an asset.

server/community_test.go
```diff
@@ -0,0 +1,69 @@
...
+					gomock.Eq(pr.Number), comment).Return(nil, nil, nil)
+
+				return client
+			},
+			OrgMembers: []string{"bar"},
+		},
+		"Org member": {
+			SetupClient: func(ctrl *gomock.Controller) *GithubClient {
+				return &GithubClient{}
+			},
```
isacikgoz, line 49: I wonder if it would makes sense to use mocks with `0` expected calls.
```suggestion
			SetupClient: func(ctrl *gomock.Controller) *GithubClient {
				issueMocks := mocks.NewMockIssuesService(ctrl)
				client := &GithubClient{
					Issues: issueMocks,
				}

				return client
			},
```

isacikgoz (approved):
Looks good to me aside from @agnivade's comments. And I have a non-blocking suggestion/question for the tests.

server/pull_request_handler.go
hanzei: Context is a good point. I haven't thought about that one.

Reverted my changes.

server/community.go
hanzei: @agnivade I was thinking about moving the test into a separate package. Does this make sense to you? If not, yes, then there is no reason to export it.

server/community_test.go
hanzei: What is the advantage? Avoid nil pointers in tests?

isacikgoz: Yes, and also makes sure that we don't receive unexpected calls. If receive any e.g. `(*GitHubClient).SendGitHubComment`, we can take a look since this test will going to fail.

server/community.go
agnivade: Having a separate test package is great, but if we have to export methods _just_ for that purpose, then we can keep it in the same package. It's more important to remain consistent IMO.

hanzei:
Based on feedback offline I've updated the PR to combine the CLA and welcome message. If the CLA is not signed, the CLA test is appended to the welcome message. On sync and PR reopening the CLA check is still performed but no message posted. This only happens when the PR is opened. Because of this behaviour change `findNeedsToSignCLAComment` is not longer needed.

Based on the discussion offline with @jasonblais and @aspleenic the welcome message was deemed useful even for existing contributors. Thanks for the suggestion @agnivade, let's keep it simple for now and post it for every PR.

I also update the code to use  `text/template` as this made the configuration easier.

@jasonblais You suggested only posting the welcome message on certain repos which follow the standard review process. It feels weird to me to not post a `Thank you ` message at all. I'm 2/5 that it's fine to start with posting the message on all repos and we can iterate if it isn't useful on certain repos.

server/community.go
hanzei: Moved to `text/template` in [`6177372` (#246)](https://github.com/mattermost/mattermost-mattermod/pull/246/commits/617737219cb8c8a3183e0907170a6b69dabd9864)

hanzei:
@agnivade @isacikgoz Please take another look

server/pull_request_handler.go
```diff
@@ -50,7 +50,10 @@ func (s *Server) pullRequestEventHandler(w http.ResponseWriter, r *http.Request)
 	switch event.Action {
 	case "opened":
 		mlog.Info("PR opened", mlog.String("repo", pr.RepoName), mlog.Int("pr", pr.Number))
-		if err = s.handleCheckCLA(ctx, pr); err != nil {
+
+		var claCommentNeeded bool
```
agnivade, line 54: Any reason why we cannot do `claCommentNeeded, err :=` ? The variable should still be available now that we have split the if condition.

server/pull_request_handler.go
```diff
@@ -71,7 +78,9 @@ func (s *Server) pullRequestEventHandler(w http.ResponseWriter, r *http.Request)
 		s.setBlockStatusForPR(ctx, pr)
 	case "reopened":
 		mlog.Info("PR reopened", mlog.String("repo", pr.RepoName), mlog.Int("pr", pr.Number))
-		if err = s.handleCheckCLA(ctx, pr); err != nil {
+
+		_, err = s.handleCheckCLA(ctx, pr)
```
agnivade, line 82: Any reason to change this? We are not using `err` anywhere after.

server/pull_request_handler.go
```diff
@@ -163,7 +172,9 @@ func (s *Server) pullRequestEventHandler(w http.ResponseWriter, r *http.Request)
 		}
 	case "synchronize":
 		mlog.Debug("PR has a new commit", mlog.String("repo", pr.RepoName), mlog.Int("pr", pr.Number))
-		if err = s.handleCheckCLA(ctx, pr); err != nil {
+
+		_, err = s.handleCheckCLA(ctx, pr)
```
agnivade, line 176: Same as above.

server/cla.go
```diff
@@ -10,16 +10,15 @@ import (
...
 	"strings"
-	"time"
 
 	"github.com/google/go-github/v32/github"
 	"github.com/mattermost/mattermost-mattermod/model"
 	"github.com/mattermost/mattermost-server/v5/mlog"
 )
 
-func (s *Server) handleCheckCLA(ctx context.Context, pr *model.PullRequest) error {
+func (s *Server) handleCheckCLA(ctx context.Context, pr *model.PullRequest) (bool, error) {
```
agnivade, line 19: Might be good to have a godoc here explaining why the boolean return is needed.

server/pull_request_handler.go
hanzei: If we do that, `err` will shadow. It's just a linter think. I personally like the `var` style more then using a different name for `err`.

server/pull_request_handler.go
hanzei: No. Reverting.

jasonblais:
@hanzei 

> You suggested only posting the welcome message on certain repos which follow the standard review process. It feels weird to me to not post a Thank you message at all. I'm 2/5 that it's fine to start with posting the message on all repos and we can iterate if it isn't useful on certain repos.

Yes, but the link to code review process doesn't apply to all repos and will be confusing. However, given staff won't receive this message, we can send it to all repos in this iteration, and adjust the exact text accordingly. I'll make the text proposals at https://github.com/mattermost/platform-private/pull/284

jasonblais (approved):

server/cla.go
hanzei: Good suggestion

server/pull_request_handler.go
agnivade: We have been using `err2` in server repo. But this is also OK.

agnivade (approved):
Thank you for your patience with this Ben! LGTM

hanzei:
Thanks for all the feedback @agnivade I've learned a lot - as always

isacikgoz (approved):
Awesome! 👍

cpanato (approved):
thanks for that!

PR #247, opened 2020-09-30
MM-28978 Move away from gorp to sqlx in issue_store.go

kashifsoofi:
#### Summary
mattermod: move away from gorp to sqlx in issue_store.go

Create types to save/load []string as json in column (as best as I could plesae let me know if i missed something)
Updated issue_store to use sqlx

#### Ticket Link
https://github.com/mattermost/mattermost-server/issues/15567

#### Jira tiket
https://mattermost.atlassian.net/browse/MM-28978


types/types.go
```diff
@@ -0,0 +1,35 @@
...
+func (j *JSONText) Scan(value interface{}) error {
+	if value == nil {
+		return nil
+	}
+
+	var source []byte
+	switch t := value.(type) {
+	case []uint8:
+		source = t
+	default:
```
agnivade, line 30: Interesting. Is this just defence-in-depth or do we really need to pass a `[]uint8` slice to `json.Unmarshal`?

model/issue.go
```diff
@@ -3,11 +3,15 @@
...
+)
+
 type Issue struct {
 	RepoOwner string
 	RepoName  string
 	Number    int
 	Username  string
 	State     string
-	Labels    []string
+	Labels    types.JSONText
```
agnivade, line 16: Perhaps `StringArray` is a better name for this? To keep the parity with `[]string`.

Let's also keep this in the model package itself.

agnivade (commented):
Apologies for the delay @kashifsoofi !

Looks really good overall. Just some small issues to take care of.

types/types.go
kashifsoofi: Copied from the sqlx example, i think as we know the type we can probably just unmarsel it without checking the type, I am kind of new to go, so i was hoping a more experienced member can comment on this :)

model/issue.go
kashifsoofi: 👍  Will update the name and location of the type

types/types.go
agnivade: Yes, I agree. This looks much better to me:

```go
func (j *JSONText) Scan(value interface{}) error {
	if value == nil {
		return nil
	}

	buf, ok := value.([]byte)
	if !ok {
		return errors.New("received value is not a byte slice")
	}

	return json.Unmarshal(buf, j)
}
```

model/StringArray.go
```diff
@@ -0,0 +1,32 @@
+package model
+
+import (
+	"database/sql/driver"
+	"encoding/json"
+	"errors"
+
+	_ "github.com/go-sql-driver/mysql" // Load MySQL Driver
```
agnivade, line 8: This is not needed here.

model/StringArray.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+type StringArray []string
```
agnivade, line 11: Let's add a godoc here explaining the purpose of this type.

model/StringArray.go
```diff
@@ -0,0 +1,32 @@
...
+
+func (sa StringArray) Value() (driver.Value, error) {
+	b, err := json.Marshal(sa)
+	if err != nil {
+		return nil, err
+	}
+	return string(b), nil
```
agnivade, line 18: Since we are getting a byte slice, it's not needed to convert to string. We can just directly do `return json.Marshal(j)`

agnivade (commented):
Looking really solid now. Just some finishing touches needed.

agnivade:
Oh, I didn't see this earlier. Conventionally, we name files in snake_case. So let's change `model/StringArray.go` to `model/string_array.go`.

agnivade (approved):
Nicely done @kashifsoofi !

isacikgoz (approved):
LGTM 👍

PR #250, opened 2020-10-08
Temporary reduce prod replicas to 1 due to pvc issue

khos2ow:
#### Summary

Currently, there are 3 replicas in production with one PVC shared with all of them. This PVC has to be of type `ReadWriteOnce` because all of the pods need to be able to write to it and as such, the volume is _bound_ to one specific node and there have been multiple `FailedScheduling` issues in the past couple of deployments. This is to reduce the replicas to 1 (for now) to make the pod _moveable_ between nodes.


metanerd (approved):
LGTM

PR #253, opened 2020-11-03
Add cloud-ga branch to need ee trigger.

metanerd:


lieut-data (approved):
Looks good! A bit weird to include changes unrelated to the core PR, but I'm not going to block the review over it.

PR #257, opened 2021-03-16
Use *bool for MaintainerCanModify 

msal4:
#### Summary
Decouple model from database-specific fields

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/16975


isacikgoz (changes requested):
Looks good @msal4 But could you please exclude `.idea/` contents? You can add these to project `.gitignore`.

mahmud2011 (changes requested):
You forgetfully committed `.idea/`

msal4:
Oops! you never saw that

util/ptr.go
```diff
@@ -0,0 +1,3 @@
+package util
+
+func Boolptr(b bool) *bool { return &b }
```
agnivade, line 3: Please move this inside the server package as an unexported function.

agnivade (changes requested):

util/ptr.go
msal4: I found out that these helper functions already exist in `utils`. I'm going to use them instead.

agnivade (approved):

isacikgoz (approved):
Thanks @msal4 🎉

PR #260, opened 2021-03-17
Dereference pointers when comparing MaintainerCanModify

msal4:
#### Summary
Right now we're comparing pointers which is probably not the behavior we want

```go
// server/pull_request_handler.go

//...
if oldPr.MaintainerCanModify == nil || (oldPr.MaintainerCanModify != pr.MaintainerCanModify) {
	prHasChanges = true
}
//...
```
I was gonna commit the change in #257 but I wasn't quick enough :)
#### Ticket Link
JIRA: https://mattermost.atlassian.net/browse/MM-26814

### DIFF
--- a/server/pull_request_handler_test.go
+++ b/server/pull_request_handler_test.go
@@ -276,8 +276,8 @@ func TestPullRequestEventHandler(t *testing.T) {
 			Times(expectedSaveCalls).Return(nil, nil)
 
 		e := pullRequestEvent{
-			Action:   "",
-			PRNumber: modelPR.Number,
+			Action:      "",
+			PRNumber:    modelPR.Number,
 			PullRequest: githubPR,
 		}
 		b, err := json.Marshal(e)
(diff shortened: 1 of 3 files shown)


agnivade (changes requested):
Woops, nice catch! Can we write a test for this please so that the behavior does not regress again?

msal4:
I think these should do it.
I'll do a separate PR after this one is merged to update `Merged, MaintainerCanModify, MilestoneNumber, MilestoneTitle` to be `*type ` since the changes are related and minuscule.

server/pull_request_handler_test.go
```diff
@@ -245,6 +245,168 @@ func TestPullRequestEventHandler(t *testing.T) {
...
+					Name:  &modelPR.RepoName,
+				},
+			},
+			Merged: &modelPR.Merged.Bool,
+			Head:   &github.PullRequestBranch{SHA: &modelPR.Sha},
+			Milestone: &github.Milestone{
+				Number: NewInt(int(modelPR.MilestoneNumber.Int64)),
+				Title:  &modelPR.MilestoneTitle.String,
+			},
+			MaintainerCanModify: NewBool(false),
```
agnivade, line 342: I'd prefer to do a `!` here too to be consistent and more robust.

agnivade (approved):

isacikgoz (approved):
LGTM 👍  Thanks for the unit tests.

PR #262, opened 2021-03-23
Decouple model from database

msal4:
#### Summary
Updates PullRequest fields to use `*type` instead of using sql nullable types.

#### Ticket Link

JIRA: https://mattermost.atlassian.net/browse/MM-26814
Fixes [#16975](https://github.com/mattermost/mattermost-server/issues/16975)


server/cherry_pick.go
```diff
@@ -92,7 +92,7 @@ func (s *Server) handleCherryPick(ctx context.Context, commenter, body string, p
 	command := getCommand(body)
 	args := strings.Split(command, " ")
 	mlog.Info("Args", mlog.String("Args", body))
-	if !pr.Merged.Valid || !pr.Merged.Bool {
+	if pr.Merged == nil || !*pr.Merged {
```
agnivade, line 95: Can we have helper methods for each of these fields? Like `IsMergedFalse`, `IsMilestoneNumberEmpty` which wraps the nil check and the false/0 check.

agnivade (changes requested):

server/cherry_pick.go
msal4: I'll name them `GetFieldName()` similar to `github.PullRequest` e.g. `GetMerged()`

agnivade (approved):
Looks so clean! Amazing work!

msal4:
@agnivade Thanks :)

isacikgoz (approved):

server/cherry_pick.go
```diff
@@ -153,8 +150,8 @@ func (s *Server) checkIfNeedCherryPick(pr *model.PullRequest) {
 	prLabels := labelsToStringArray(labels)
 	for _, prLabel := range prLabels {
 		if prLabel == "CherryPick/Approved" {
-			milestoneNumber := int(pr.MilestoneNumber.Int64)
-			milestone := getMilestone(pr.MilestoneTitle.String)
+			milestoneNumber := int(*pr.MilestoneNumber)
```
isacikgoz, line 153:
```suggestion
			milestoneNumber := int(pr.GetMilestoneNumber())
```

model/pull_request.go
```diff
@@ -28,9 +27,37 @@ type PullRequest struct {
...
 	MergeCommitSHA      string `db:"-"`
-	MaintainerCanModify sql.NullBool
-	MilestoneNumber     sql.NullInt64
-	MilestoneTitle      sql.NullString
+	MaintainerCanModify *bool
+	MilestoneNumber     *int64
+	MilestoneTitle      *string
+}
+
+func (pr *PullRequest) GetMerged() bool {
```
isacikgoz, line 37: Not a blocking thing but, I'd prefer Go convention:
```suggestion
func (pr *PullRequest) Merged() bool {
```

isacikgoz (approved):
Just a couple of non blocking points.

model/pull_request.go
msal4: @isacikgoz Wouldn't that conflict with the field name?

agnivade: Ahh yeah .. that's true, the field is exported. Well, we could unexport and have it accessible only via a getter. But it's out of scope for this PR.

msal4: I'll happily do a separate PR addressing it if I get a green light.

agnivade: I don't think that's necessary, just for the sake of sticking to a convention. A comment explaining why the "Get" part is needed should suffice.

PR #263, opened 2021-03-25
[MM-33617] Assign community greeter and common labels

Willyfrog:
#### Summary

When a new community PR is created, assign randomly a member of a dedicated team for that repo to do an initial assesment of the PR (assign an additional reviewer, add labels as they feel are needed, etc.)

it also allows to define labels to be applied as the most common. In our use case those would be `Dev Review` and `QA Review`, leaving to the greeter member to decide on others that might be needed as well (Product, UX, Security, etc.)

#### Ticket Link

[MM-33617](https://mattermost.atlassian.net/browse/MM-33617)

#### Additional Notes

once this PR is merged, we'll need to change the configuration for webapp and desktop teams to include the options mentioned in the ticket.


server/config.go
```diff
@@ -36,6 +36,8 @@ type Repository struct {
 	InstanceSetupScript        string
 	InstanceSetupUpgradeScript string
 	JobName                    string
+	GreetingTeam               string // Team of staff who is in charge of managing new community PRs for this repo
```
lieut-data, line 39:
```suggestion
        // GreetingTeam is the GitHub team responsible for triaging non-member PRs for this repo.
	GreetingTeam               string 
```

server/config.go
```diff
@@ -36,6 +36,8 @@ type Repository struct {
...
+	GreetingLabels             []string
```
lieut-data, line 40:
```suggestion
	// GreetingLabels are the labels applied automatically to non-member PRs for this repo.
	GreetingLabels             []string
```

server/config.go
lieut-data: Proposing `non-member` to highlight the distinction being made in not applying this process to members of the repository.

server/community_test.go
```diff
@@ -93,3 +93,214 @@ func TestPostPRWelcomeMessage(t *testing.T) {
...
+			GreetingLabels: []string{"hey", "hello"},
+		}
+
+		s := &Server{
+			Config: &Config{
+				Repositories: []*Repository{repo},
+			},
+			OrgMembers: []string{"foo"},
+		}
+		assert.NoError(t, s.assignCommunityLabels(context.Background(), pr, repo))
```
lieut-data, line 118: I'm guessing the absence of `EXPECT` assertions here is sufficient to verify that we /aren't/ doing anything for org members in this case. Is there any way with gomock to make that explicit for clarity in the test case itself?

server/community_test.go
```diff
@@ -93,3 +93,214 @@ func TestPostPRWelcomeMessage(t *testing.T) {
...
+
+	t.Run("Happy path", func(t *testing.T) {
+		repo := &Repository{
+			Owner:        "owner",
+			Name:         "repoName",
+			GreetingTeam: "greetingTeam",
+		}
+		userLogin := "bar"
+		greeter := &github.User{Login: &userLogin}
+		members := []*github.User{greeter}
```
lieut-data, line 260: Can we test the "randomness" of having multiple users to somehow?

server/community.go
```diff
@@ -64,3 +66,62 @@ func (s *Server) postPRWelcomeMessage(ctx context.Context, pr *model.PullRequest
 
 	return nil
 }
+
+func (s *Server) assignGreeter(ctx context.Context, pr *model.PullRequest, repo *Repository) error {
+	// Only assign an Usher for community PRs
```
lieut-data, line 71:
```suggestion
	// Only assign an greeter for community PRs
```

server/community.go
```diff
@@ -64,3 +66,62 @@ func (s *Server) postPRWelcomeMessage(ctx context.Context, pr *model.PullRequest
...
+	if s.IsOrgMember(pr.Username) {
+		return nil
+	}
+
+	// Is it configured to have a greeting team to handle the PR?
```
lieut-data, line 76: Should we test this first to avoid unnecessary GitHub client API calls for `IsOrgMember` when a repo has no such configuration?

lieut-data (changes requested):
Oh man, this is going to be awesome! Love that this is general purpose allowing optional adoption by other teams -- nice work, @Willyfrog!

A few comments below.

server/community.go
metanerd: IsOrgMember fetches list of org members in a CRON once per day, so this shouldn't be an issue.

server/community.go
Willyfrog: haha, that was the original name, but I didn't like it for looking too similar to user, which might end up in people misreading later on. thanks for catching that one.

server/community_test.go
Willyfrog: I'll explicitly add that, but since the `server` doesn't have a githubclient configured it would fail

server/community_test.go
Willyfrog: since it is using a core function, I don't think we need to test that, but if we go into some other algorithm (round robin, balanced, etc.) we should definitely test it.

lieut-data (approved):

lieut-data:
@metanerd, what's involved in changing the configuration for this after being merged? Is that in a separate repo?

server/community.go
```diff
@@ -64,3 +66,62 @@ func (s *Server) postPRWelcomeMessage(ctx context.Context, pr *model.PullRequest
...
+	if repo.GreetingTeam == "" {
+		return nil
+	}
+
+	// get available members
+	greeterList, _, err := s.GithubClient.Teams.ListTeamMembersBySlug(ctx, s.Config.Org, repo.GreetingTeam, nil)
+	if err != nil {
+		mlog.Error("Error retrieving greeting team members", mlog.Err(err), mlog.String("GreetingTeam", repo.GreetingTeam))
```
agnivade, line 84: We can remove this and avoid the double-logging, and add context to the error instead.

Also, we can remove the word "greeting" because that context is already there in the top-level log statement, this is just about the GitHub call failing.

server/community.go
```diff
@@ -64,3 +66,62 @@ func (s *Server) postPRWelcomeMessage(ctx context.Context, pr *model.PullRequest
...
+	}
+
+	greeter := greeterList[rand.Intn(size)]
+	greetingRequest := github.ReviewersRequest{
+		Reviewers: []string{*greeter.Login},
+	}
+
+	_, _, err = s.GithubClient.PullRequests.RequestReviewers(ctx, pr.RepoOwner, pr.RepoName, pr.Number, greetingRequest)
+	if err != nil {
+		mlog.Warn("Couldn't assing a member of the greeting team", mlog.Err(err), mlog.String("GreetingTeam", repo.GreetingTeam))
```
agnivade, line 102: Any reason not to return the error here? This seems like a failure, and we should just return the error with context and let the upper layer log it.

server/community.go
```diff
@@ -64,3 +66,62 @@ func (s *Server) postPRWelcomeMessage(ctx context.Context, pr *model.PullRequest
...
+	greeterList, _, err := s.GithubClient.Teams.ListTeamMembersBySlug(ctx, s.Config.Org, repo.GreetingTeam, nil)
+	if err != nil {
+		mlog.Error("Error retrieving greeting team members", mlog.Err(err), mlog.String("GreetingTeam", repo.GreetingTeam))
+		return err
+	}
+
+	// Assign one of them as a reviewer
+	size := len(greeterList)
+	if size == 0 {
+		mlog.Warn("There were no members on the greeting team", mlog.String("GreetingTeam", repo.GreetingTeam))
```
agnivade, line 91: I would prefer this to be an error condition, so that the app is forced to run correctly. Unless there is a scenario where it's valid to have a team with no members?

server/community.go
```diff
@@ -64,3 +66,62 @@ func (s *Server) postPRWelcomeMessage(ctx context.Context, pr *model.PullRequest
...
+
+	// Is repo configured to setup labels for non-member PRs?
+	if len(repo.GreetingLabels) == 0 {
+		return nil
+	}
+
+	// Assign labels
+	_, _, err := s.GithubClient.Issues.AddLabelsToIssue(ctx, pr.RepoOwner, pr.RepoName, pr.Number, repo.GreetingLabels)
+	if err != nil {
+		mlog.Error("Error applying Greeting labels", mlog.Err(err), mlog.Int("PR", pr.Number), mlog.String("Repo", pr.RepoName))
```
agnivade, line 123: Same as above.

server/community.go
```diff
@@ -64,3 +66,62 @@ func (s *Server) postPRWelcomeMessage(ctx context.Context, pr *model.PullRequest
...
+		return nil
+	}
+
+	// Is repo configured to setup labels for non-member PRs?
+	if len(repo.GreetingLabels) == 0 {
+		return nil
+	}
+
+	// Assign labels
+	_, _, err := s.GithubClient.Issues.AddLabelsToIssue(ctx, pr.RepoOwner, pr.RepoName, pr.Number, repo.GreetingLabels)
```
agnivade, line 121: Does this handle the case where the labels are already attached to the issue? For example, I use the `gh` CLI tool to automatically assign labels and reviewers when I create a PR. Some contributors might also do the same.

agnivade (changes requested):
Thanks @Willyfrog ! Just some minor improvements to take care of.

And in case you haven't already, I would suggest to test this out manually first with the https://github.com/mattermosttest/mattermost-server repo. The instructions to setup ngrok is there in this repo's [README](https://github.com/mattermost/mattermost-mattermod#mattermod-local-testing).

Willyfrog:
> And in case you haven't already, I would suggest to test this out manually first with the https://github.com/mattermosttest/mattermost-server repo. The instructions to setup ngrok is there in this repo's [README](https://github.com/mattermost/mattermost-mattermod#mattermod-local-testing).

I tested this, but with a testing org I created for that purpose

> Does this handle the case where the labels are already attached to the issue? For example, I use the gh CLI tool to automatically assign labels and reviewers when I create a PR. Some contributors might also do the same.

I tried running the same payload, and it didn't break. But I'll try with other labels as well just to make sure it didn't do anything weird

server/community.go
```diff
@@ -64,3 +67,58 @@ func (s *Server) postPRWelcomeMessage(ctx context.Context, pr *model.PullRequest
...
+
+	// Is it configured to have a greeting team to handle the PR?
+	if repo.GreetingTeam == "" {
+		return nil
+	}
+
+	// get available members
+	greeterList, _, err := s.GithubClient.Teams.ListTeamMembersBySlug(ctx, s.Config.Org, repo.GreetingTeam, nil)
+	if err != nil {
+		return errors.Wrapf(err, "Error retrieving team members from %s", repo.GreetingTeam)
```
agnivade, line 85: Nit: error messages should not be capitalized: https://github.com/golang/go/wiki/CodeReviewComments#error-strings. (Here and below too)

agnivade (approved):
Looks good. Please also take a look at the linter issues.

PR #268, opened 2021-05-05
MM-36618: switch to assigning whole greeter team

lieut-data:
#### Summary
Instead of querying the configured `GreetingTeam`s members and assigning one at random, just assign the whole team, and rely on the GitHub team having been configured to auto-assign accordingly:

image

This approach automatically gets out "Out of office" support, since GitHub wont' assign a team member who's out of office. There's also a bunch of nice configuration options on GitHub, including the ability to restrict assignees to a subset of team members, and pick how many should be assigned.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-33618

### DIFF
--- a/server/community.go
+++ b/server/community.go
@@ -6,8 +6,6 @@ package server
 import (
 	"bytes"
 	"context"
-	"fmt"
-	"math/rand"
 	"text/template"
 	"time"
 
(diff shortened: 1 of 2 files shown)


metanerd (approved):
LG AFAIU
Less logic to maintain.

PR #269, opened 2021-06-02
WIP Add ok to test feature 

cpanato:
#### Summary
Add ok to test feature.

How this will works:

Having a CircleCI job configured we have the following cases

**PR opened in upstream**
- it will run the CircleCI job normally as similar we have today

**PR opened from a forked repository but the user is a Mattermost Org member**
- It will trigger the CircleCI Job using the mattermod user to inject the necessary secrets, similar we have today, but if that touches the block list files it will just alert but will continue the CI run. Because it came from a trusted org member

**PR opened from a forked repository but the user is NOT a Mattermost Org member**
- CircleCI trigger will be blocked and will not run
- A `Need Ok to Test` label will be added
- A GitHub Status will be added saying the PR is blocked to be merged
- A message will be added, like
```
Hi {{.Username}}. Thanks for your pull request!\n\nI'm waiting for a Core Committer will review your pull request soon to verify that this patch is reasonable to test. If it is, they should reply with `/ok-to-test` on its own line or add the `Ok to Test` label. Until that is done, I will not automatically test new commits in this PR.\n\nOnce the patch is verified, the new status will be reflected by the `Ok to Test` label
```

A Core Committer will review the PR and if that sounds reasonable to test, can add the `Ok to Test` label and that will trigger the Circleci jobs and will remove the `Need Ok to Test` and the GitHub status blocker.

If the contributor pushes new commits the label `Ok to Test` is still in place it will trigger new CI jobs. If someone removes the label and the contributor pushes new commits it will block the CI job and re-add the label `Need Ok to Test`

This helps with the security concerns about leaking possible secrets.
Other security improvements will come soon, like reduce the S3/Docker access. 

A demo was shown at the Developer Meeting some weeks ago.

Opened the PR to gather feedback while I'm implementing the slash-command parser

## TODO:
- Add the parser `/ok-to-test` command

#### Ticket Link
JIRA: https://mattermost.atlassian.net/browse/DOPS-430


cpanato:
cc @jwilander @DSchalla

cpanato:
When this is ok to be merged I will update the configuration

agnivade:
Haven't taken a deeper look, but just a quick question: curious to know why we are creating a separate workflow when we already have a `EEtests` label. The current EEtests label will not run the enterprise tests for any contributor PR. It seems like we are just expanding the scope of that to not run _any_ tests at all.

I would expect the label to be renamed to something else (perhaps `run-tests` 0/5), and it runs all the tests keeping all the other functionality same.

The `eetests` label has a slightly stricter workflow where it always removes the label every time someone adds it. This prevents the contributor from pushing bad code _after_ an org member has added the label which will always persist. It's a bit more laborious, but the Go project currently uses the exact same workflow and given that we already do the same for EE tests, just changing the label and expanding the workflow to run all tests shouldn't add any extra overhead.

Thoughts?

cpanato:
@agnivade thanks for the question

The idea is to block any tests that are coming from a fork and non-org member, this was discussed in a dev meeting some time ago and also address some security concerns from the Sec Team. The EEtests is one layer of thing we have to avoid exposing the ee code, but in the  TE pipeline, we have other sensitive data that can be exposed, like AWS keys/docker hub and other API keys.

But we can remove the EE label and have the ok to test label, when that is added we can trigger the EE as well if that comes from the mattermost-server, if not it will just trigger the normal pipeline.

Also, I can adjust the case to remove the label after that is triggered to avoid the new push with bad code. 

let me know if this approach sounds good to you and I will update the code accordinling

agnivade:
> But we can remove the EE label and have the ok to test label, when that is added we can trigger the EE as well 

> Also, I can adjust the case to remove the label after that is triggered to avoid the new push with bad code.

Yes, that is effectively replacing the "EEtests" label with a new label. That's exactly what I was suggesting. The proposed changes sounds good to me. Thanks Carlos.

metanerd:
My biggest concern is that if we want to actively depend on it, we need to fix the rate limit issue first: https://community-daily.mattermost.com/private-core/pl/74xripxnufrpmfa9ojsjfy6rbw

PR #272, opened 2021-09-23
Add contributor label to all PRs

puerco:
#### Summary

This PR modifies the bot to add the contributor label _Contributor_ to all PRs and
issues filed by authors who are not in the org.

**Please note that this should not be merged until the label exists in all repos.**

Signed-off-by: Adolfo García Veytia (Puerco) <adolfo.garcia@mattermost.com>

#### Ticket Link

https://mattermost.atlassian.net/browse/DOPS-262

### DIFF
--- a/server/community.go
+++ b/server/community.go
@@ -15,6 +15,8 @@ import (
 	"github.com/pkg/errors"
 )
 
+const contributorLabel = "Contributor"
+
 func (s *Server) addHacktoberfestLabel(ctx context.Context, pr *model.PullRequest) {
 	if pr.State == model.StateClosed {
 		return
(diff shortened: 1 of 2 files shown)


server/community.go
```diff
@@ -95,14 +97,25 @@ func (s *Server) assignGreetingLabels(ctx context.Context, pr *model.PullRequest
 		return nil
 	}
 
-	// Is repo configured to setup labels for non-member PRs?
-	if len(repo.GreetingLabels) == 0 {
-		return nil
+	// Check if the repository has the contriutor label already
```
cpanato, line 100: s/contriutor/contributor

server/community.go
```diff
@@ -95,14 +97,25 @@ func (s *Server) assignGreetingLabels(ctx context.Context, pr *model.PullRequest
...
+	labels := repo.GreetingLabels
```
cpanato, line 101: this is a config for each repo, this does not come from the PR.

I think in this function we need to check if the PR submitter is a member of the org and if not we add the label. and ignore if the `GreetingLabels` is configured or not.

WDTY?

cpanato (changes requested):
one nit typo and a comment


thanks for being quick on pick the issue :)

server/community.go
puerco: Yep it is checked (line 96). But instead of ignoring the GreetinLabels, what the patch does is append the `Contributor` label to the []GreetingLabels that the repo already has configured. 

TBH, it does not feel 100% right to hardcode the addition of the label to all PRs. I'd like to have control if I don't want this happening on some repos. Perhaps it could be a flag in the repo configuration?

server/community.go
puerco: Done

server/community.go
cpanato: all good! thanks for the clarification

cpanato (commented):
add this in the line 73
```
if s.IsOrgMember(pr.Username) || s.IsBotUserFromCLAExclusionsList(pr.Username) {
    return nil
}
```

puerco:
The bot change is :)

cpanato (approved):
thanks!

PR #277, opened 2021-10-12
Do not review mlog changes for organization members

isacikgoz:
#### Summary

I think everyone should be aware of the logging guidelines so far. Let's review this for non-org members.


server/review.go
```diff
@@ -20,6 +20,12 @@ import (
 const mlogReviewCommentBody = "Gentle reminder to check our logging [principles](https://developers.mattermost.com/contribute/server/style-guide/#log-levels) before merging this change."
 
 func (s *Server) reviewMlog(ctx context.Context, pr *model.PullRequest, nodeID, diffURL string) error {
+	for _, m := range s.OrgMembers {
```
cpanato, line 23: i think you can use `s.IsOrgMember(pr.Username)`

isacikgoz: Thanks for pointing that one, I also added a comment.

cpanato (approved):
will rollout this tomorrow

PR #279, opened 2021-10-20
use team reviewers instead of reviewers

cpanato:
#### Summary
in the mattermod config we have setup a team and not individuals, so the API call failed because that is not a unique member and it is a team

#### Ticket Link
JIRA: https://mattermost.atlassian.net/browse/DOPS-669

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,17 +6,16 @@ orbs:
 executors:
   default:
     docker:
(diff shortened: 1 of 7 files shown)


lieut-data (approved):
Sweet! Thank you :)

hmhealey (approved):

cpanato:
deployed

PR #280, opened 2021-10-21
New golang based CP logic and /cherry-pick beta command

puerco:
#### Summary

This PR adds the new golang-based cherry-pick logic to mattermod to support pull requests merged in any fashion: squashed, merge-commit or rebased. Before this change, mattermod would only cherry pick PRs which were merged by squashing. 

This PR also modifies the mattermod `cherry-pick` command to call the new logic in beta mode. To use the new object, add `beta` to the command:

```
/cherry-pick beta branchName
``` 

Once the new logic is tested and stable we should remove the beta command and the cherry-pick script.


puerco:
@cpanato we need to bump the `golangci-lint` image as well because to go version that ships with it is a bit outdated:
```
common.go:439:19: ReadDir not declared by package os
```

It is still at 1.27:
https://github.com/mattermost/mattermost-mattermod/blob/e80ab498c2c5596636f8bf453a36d99d114533ce/.circleci/config.yml#L11-L18

I'll take care of it tomorrow.

agnivade:
@isacikgoz worked on the cherry-pick implementation. He would be a better person to look at this.

server/cherrypicker.go
```diff
@@ -0,0 +1,571 @@
...
+	}
+
+	// Open the repo
+	repo, err := git.PlainOpen(opts.RepoPath)
+	if err != nil {
+		return errors.Wrapf(err, "opening repository from %s", opts.RepoPath)
+	}
+
+	// And add it to the state
+	state.Repository = repo
```
isacikgoz, line 128: Is this used anywhere? `go-git` is a big library and I couldn't understand why we would need this?

isacikgoz (commented):
Wow too much effort here @puerco 🚀 Sorry with the late review as I have a long queue 😅 I see that the Go based library `go-git` is not used as much. There are many `gitCommand` usage. What is the plan for the future? I am a bit confused about how the transition will take place. On the other hand, I have some other comments below (or above).

puerco:
Hey @isacikgoz Thanks for the thorough review!
We are starting to split out some of the code in mattermod into independent packages to reuse between the new pipeline code we are writing and the new mattermod iteration.

The cherry-pick code in this PR now lives over here:  https://github.com/mattermost/cicd-sdk/tree/master/pkg/cherrypicker

I will open a PR on the new repo addressing your comments, thanks a lot!!

cpanato:
@puerco any updates on this?

PR #283, opened 2021-11-23
Extend polling interval to 30 secs instead of 5 secs

metanerd:
#### Summary
 The waiting for a GitHub status polling stresses the secondary rate limit https://docs.github.com/en/rest/guides/best-practices-for-integrators#dealing-with-secondary-rate-limits.
Until we implement the `Retry-After` into the GitHubClientWithLimiter, let's decrease the stress.

### DIFF
--- a/server/github.go
+++ b/server/github.go
@@ -332,7 +332,7 @@ func (s *Server) createRepoStatus(ctx context.Context, pr *model.PullRequest, st
 }
 
 func (s *Server) waitForStatus(ctx context.Context, pr *model.PullRequest, statusContext string, statusState string) error {
-	ticker := time.NewTicker(5 * time.Second)
+	ticker := time.NewTicker(30 * time.Second) // extended to 30 seconds to go with the secondary rate limit interval of GitHub
 	for {
 		select {
 		case <-ctx.Done():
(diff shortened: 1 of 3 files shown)


server/github.go
```diff
@@ -332,7 +332,7 @@ func (s *Server) createRepoStatus(ctx context.Context, pr *model.PullRequest, st
 }
 
 func (s *Server) waitForStatus(ctx context.Context, pr *model.PullRequest, statusContext string, statusState string) error {
-	ticker := time.NewTicker(5 * time.Second)
+	ticker := time.NewTicker(30 * time.Second) // extended to 30 seconds to go with the secondary rate limit interval of GitHub
```
saturninoabril, line 335: It's a big jump. I think this makes sense with E2E but I don't know if we want that for other case like triggering EE build where the user somehow expects an "immediate" response.

saturninoabril (approved):
Thanks Elisabeth! I agree that the better approach would be the `Retry-After`.

server/github.go
metanerd: Maybe I should hand ticker time in as a parameter?

saturninoabril: Sounds good to me and thanks for considering!

metanerd: Of course!

saturninoabril: Thanks! LGTM.

metanerd:
@cpanato gentle reminder to review.

metanerd:
@cpanato gentle ping on this.

cpanato (approved):

cpanato:
deployed

PR #299, opened 2022-01-26
Fix e2e test comment spam

metanerd:
Fixes https://mattermost.atlassian.net/browse/DOPS-759 :crossed_fingers: 
Depends on https://git.internal.mattermost.com/gitlab/argo-shared-cluster/-/merge_requests/9

#### Ticket Link
https://mattermost.atlassian.net/browse/DOPS-759


Phrynobatrachus:
need a bit more time to review but I did notice something kind of similar in:

https://github.com/mattermost/mattermost-webapp/pull/9715

I put the `release-note-none` label on but I didn't annote the `NONE` in the post content with `release-note`, so the bot ended up posting 3 times about it.

server/e2e_test_command.go
```diff
@@ -15,7 +15,6 @@ import (
 const (
 	e2eTestMsgCommenterPermission = "You don't have permissions to trigger this command.\n It's only available for organization members."
 	e2eTestMsgCIFailing           = "The command /e2e-test requires all PR checks to pass."
```
saturninoabril, line 17: Is there no potential cyclic issue when this message is commented my mattermod? I wonder if this message could be made more generic like "The slash command requires all PR checks to pass.". Though it might not be obvious which slash command it's referring to.

server/e2e_test_command.go
```diff
@@ -249,3 +232,22 @@ func (s *Server) getBranchAndSHAWithSameName(ctx context.Context, owner string,
...
+func (s *Server) waitForImageStatus(pr *model.PullRequest) (bool, error) {
+	ghStatusContext := ""
+	switch pr.RepoName {
+	case s.Config.E2EWebappReponame:
+		ghStatusContext = s.Config.E2EWebappStatusContext
+	case s.Config.E2EServerReponame:
+		ghStatusContext = s.Config.E2EServerStatusContext
+	}
+
+	ctx2, cancel := context.WithDeadline(context.Background(), time.Now().Add(s.Config.E2ETestDeadline))
```
saturninoabril, line 245: The value of `s.Config.E2ETestDeadline)` at the deployment is 7200 which I believe is 2 minutes. Isn't it very long duration?

saturninoabril (commented):
Thanks Elisabeth! All looks good to me. Left couple of comments just to clarify some items.

server/e2e_test_command.go
metanerd: Oh my, yes, this is what is causing it :face_palm: ! Thanks! 
Still doesn't explain why the tests didn't catch it. I will fix this.

server/e2e_test_command.go
metanerd: I wanted to give it enough time to be able to wait for building image. I am fine with something shorter though. What do you think would be a good time? 5 mins polling every 30 secs?

saturninoabril: Ah, I'm not correct, 7200 is 2 hours. Yup, 5 mins polling every 30 secs sounds good to me. Then, let the user read the message that the image is not yet available, and then do the retry manually as they want to.

Phrynobatrachus (approved):
LGTM, thanks!

saturninoabril (approved):
Thanks @metanerd! LGTM.

metanerd:
Kind ping on this @spirosoik .

spirosoik (approved):
lgtm

PR #300, opened 2022-03-24
Fix polling for github status for e2e testing.

metanerd:
#### Summary
I merged this PR (https://github.com/mattermost/mattermost-mattermod/pull/299/files) recently to fix the e2e-test spam messages, when a PR is not in a ready state. 
Turns out that polling now for the CI status is completely broken. 
Honestly, I am guessing on what the fix is. The answer is immediate, there is no attempt or log message to even try to poll. It might be that I tried to use the `context.WithDeadline` compared to what we usually use is `context.WithTimeout`. I also unnecessarily changed getting instead of the list of statuses for the sha to ref, which I also revert with this PR.

### DIFF
--- a/server/e2e_test_command.go
+++ b/server/e2e_test_command.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"net/http"
 	"strings"
-	"time"
 
 	"github.com/mattermost/mattermost-mattermod/model"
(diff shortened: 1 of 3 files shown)


saturninoabril (approved):
Thanks Elisabeth!

spirosoik (approved):
Thanks :)

PR #301, opened 2022-03-24
Revert using combined status instead of specific github status context.

metanerd:
#### Summary
 This is a partial revert, fixing the e2e-spam, but not using the specific GitHub status context to only require docker build steps to pass. The whole CI pipeline still needs to pass.


saturninoabril (approved):
Thanks Elisabeth!

PR #304, opened 2022-03-31
[MM-42724] Trigger E2E tests through PR labeling

mvitale1989:
#### Summary

This PR adds the possibility to trigger E2E tests by labeling the PR. The label value is configurable in the `E2ETriggerLabel` config key, which in the config sample template is set to `QA review/deferred`.
This is in addition to the usual trigger through the `/e2e-test` comment text, which continues working.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/19829


metanerd:
Thank you for the PR @mvitale1989 !

server/e2e_test_qa_workflow.go
```diff
@@ -0,0 +1,101 @@
...
+			mlog.Err(e2eTestQAErr))
+		return
+	}
+
+	mlog.Debug("Determined that the event should trigger the E2E test",
+		mlog.Int("pr", pr.Number),
+		mlog.String("repo", pr.RepoName))
+	err = s.handleE2ETest(ctx, s.Config.Username, pr, "")
+	if err != nil {
+		mlog.Error("Error in triggering the E2E test from PR event",
```
saturninoabril, line 80:
```suggestion
		mlog.Error("Error in triggering the E2E test from PR's label event",
```

server/e2e_test_qa_workflow.go
```diff
@@ -0,0 +1,101 @@
...
+			mlog.Int("pr", pr.Number),
+			mlog.String("repo", pr.RepoName),
+			mlog.Err(err))
+	}
+}
+
+func hasMoreThanOneApproval(reviews []*github.PullRequestReview) bool {
```
saturninoabril, line 87: The naming seemed to be not accurate since the for loop returns true for atleast one "approved" state.

server/e2e_test_qa_workflow.go
```diff
@@ -0,0 +1,101 @@
...
+	for _, review := range reviews {
+		if *review.State == "approved" {
```
saturninoabril, line 89: I've seen "approved" in another place (corresponding test). Should we make it as constant?

saturninoabril (changes requested):
Thanks @[mvitale1989](https://github.com/mvitale1989). Overall looks good to me! Please see comments.

config/config-mattermod.default.json
```diff
@@ -66,6 +66,7 @@
     "E2EServerReponame": "",
     "E2EWebappStatusContext": "",
     "E2EServerStatusContext": "",
+    "E2ETriggerLabel": "QA review/deferred",
```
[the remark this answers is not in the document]
metanerd: Agreed, we can reuse the existing labels. This config is a dummy config. The real config is internally. I will add you as reviewer for the real config, before we deploy this change.

server/e2e_test_qa_workflow.go
```diff
@@ -0,0 +1,101 @@
+package server
```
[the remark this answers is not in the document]
metanerd: I am fine with renaming this. Naming is hard.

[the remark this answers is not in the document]
mvitale1989: Sounds good to me, renamed to `server/e2e_test_from_label.go` as suggested

server/e2e_test_qa_workflow.go
```diff
@@ -0,0 +1,101 @@
...
+	default:
+		panic("unhandled error type")
+	}
+}
+
+type E2ETestQAError struct {
+	source string
+}
+
+func (s *Server) triggerE2ETestFromPRChange(ctx context.Context, pr *model.PullRequest) {
```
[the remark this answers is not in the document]
mvitale1989: sgtm, consistent with the `TestFromLabel` suffixing. Renamed

server/e2e_test_qa_workflow.go
mvitale1989: totally right, renamed to `hasAtLeastOneApproval`

server/e2e_test_qa_workflow.go
mvitale1989: Done

server/e2e_test_qa_workflow.go
```diff
@@ -0,0 +1,101 @@
...
+	case e2eTestQAMsgPRNotMergeable:
+		return "e2e not running, pr not in mergeable state"
+	case e2eTestQAMsgPRHasNoApprovals:
+		return "e2e not running, pr has no approval"
+	default:
+		panic("unhandled error type")
+	}
+}
+
+type E2ETestQAError struct {
```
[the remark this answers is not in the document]
mvitale1989: Makes sense, changed all instances of `QA*` to `FromLabel*`

mvitale1989:
Thank you for the suggestions, I applied them locally to also update other related lines.
However i also noticed a problem in the reliability of the tests, after the move to a separate file. Will fix tomorrow

mvitale1989:
Fixed. 
The issue was a race condition about the shared usage of the `msg` variable across test cases: both the testing framework and the HTTP server implementations use goroutines, but they're not always synchronized it seems, so the changing of `msg` in a test case could happen _before_ the previous test case's HTTP request handling was completed, which leads to occasional failed assertions (depends on which goroutine wins the race).

server/e2e_test_from_label.go
```diff
@@ -0,0 +1,102 @@
...
+			mlog.Err(err))
+		return
+	}
+	if !hasAtLeastOneApproval(prReviews) {
+		e2eTestFromLabelErr = &E2ETestFromLabelError{source: e2eTestFromLabelMsgPRHasNoApprovals}
+		mlog.Warn("Not triggering E2E test, due to missing required approvals",
+			mlog.Int("pr", pr.Number),
+			mlog.String("repo", pr.RepoName),
+			mlog.Err(e2eTestFromLabelErr))
+		return
```
saturninoabril, line 55: @mvitale1989 @metanerd How are we going to deal with error which caused not to run E2E automatically so that a PR will not miss E2E run before merge?

saturninoabril (commented):
Thanks Mario for updating!

server/e2e_test_from_label.go
metanerd: We are not dealing with that at the moment. Let's deal later with it, when we see it happening? I have an idea, but we have to be able to fully rely on our infrastructure for the tests, otherwise teams will complain to not being able to merge.

saturninoabril (approved):
Thanks @mvitale1989! Looks good to me and looking forward to see this in action soon.

Phrynobatrachus (approved):
Cool! I haven't fully done a local setup but tests and code overall LGTM, don't want to block this extra time. Thanks!

PR #305, opened 2022-04-07
server/issues: add automated fast-forward process

isacikgoz:
#### Summary
The following repositories needs to be added to the config:
```json
{
	"CloudRepositories": [
		"mattermost-server",
		"mattermost-webapp",
		"enterprise",
		"mattermost-api-reference"
	]
}
```

The slash command would be only executed from a mattermost-server issue and it needs to be done by a staff member.


isacikgoz:
@agnivade This was so much fun :p

server/fast_forward.go
```diff
@@ -0,0 +1,123 @@
...
+
+	"github.com/google/go-github/v39/github"
+	"github.com/mattermost/mattermost-mattermod/model"
+	"github.com/mattermost/mattermost-server/v5/mlog"
+)
+
+const (
+	cloudBranchName = "heads/cloud"
+	mainBranchName  = "heads/master"
+	processRepo     = "mattermost-server"
```
agnivade, line 20: Hmm .. curious to know about the name "processRepo"? I'd expect it to be validRepo or sourceRepo. 0/5 though. Just curious :)

server/fast_forward.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	// Don't start process if the user is not a core committer
+	if !s.IsOrgMember(user) {
+		return nil
+	}
+
+	// Check if the args are correct
+	command := getFFCommand(comment)
+	args := strings.Split(command, " ")
```
agnivade, line 40: Can we not do the arg generation automatically based on the current date? I'd prefer it to be as hassle-free as possible.

agnivade (commented):
Looks like a great start :) Couple of comments:

1. We would want to process this whole thing asynchronously. GitHub webhooks expect to get 
the response fast, and any long processing needs to be done asynchronously.
2. We need to protect against concurrent `/cloud-ff` commands. If an existing command is already in progress, throw an error.
3. An important feature for this is that we want the command to be idempotent. That way, even if someone runs the command twice in the same week, it should detect that a branch already exists and return appropriately. That's why I was suggesting not to pass the date, and let the tool automatically deduce the date and first check if a branch already exists with the same name, and if so, return an error. Makes things more simple and bulletproof design.
4. Instead of posting multiple comments, let's post a single big comment containing a record of everything we did or failed to do.
   - Created backup branch
   - Deleted cloud branch
   - Created new cloud branch

isacikgoz:
> * We would want to process this whole thing asynchronously. GitHub webhooks expect to get
>   the response fast, and any long processing needs to be done asynchronously.

Do you mean asynchronously by executing the process for each repository concurrently? Since these branches are not created from scratch, it's actually pretty instantaneous. We are just adding a ref to a commit and deleting the ref. In fact, doing this concurrently maybe make us hit the rate limiter.

> * We need to protect against concurrent `/cloud-ff` commands. If an existing command is already in progress, throw an error.

Sure, I wouldn't expect this to happen pretty often but for a peace of mind, we can guard this with a mutex. Considering we are running mattermod as a single instance.

> * An important feature for this is that we want the command to be idempotent. That way, even if someone runs the command twice in the same week, it should detect that a branch already exists and return appropriately. That's why I was suggesting not to pass the date, and let the tool automatically deduce the date and first check if a branch already exists with the same name, and if so, return an error. Makes things more simple and bulletproof design.

We can do that, in fact we can improve this with a 5 day check to the last backup, if exist (we assume that ff process is executed in the past 5 days) report the issue and ask for a `--force` flag to ignore this constraint. WDYT?
 
> * Instead of posting multiple comments, let's post a single big comment containing a record of everything we did or failed to do.

Sure, makes sense.

agnivade:
> Do you mean asynchronously by executing the process for each repository concurrently?

Ah no no, just do the whole thing concurrently. As in `go performFastForwardProcess()`. But if you are saying it's fast, then let's let it be. :+1: 

> Sure, I wouldn't expect this to happen pretty often but for a peace of mind, we can guard this with a mutex. Considering we are running mattermod as a single instance.

Good point, actually mattermod runs in a 2 pod replica setup in k8s. So we'd need to have a DB flag for this. But since we are doing the whole thing synchronously, it shouldn't be that hard. Just set a flag while entering the function, and clear it in a defer clause. And before entering, check if the flag is set.

Since this is directly changing the prod branches, let's take the extra step to not allow the tool to be misused.

> We can do that, in fact we can improve this with a 5 day check to the last backup, if exist (we assume that ff process is executed in the past 5 days) report the issue and ask for a --force flag to ignore this constraint. WDYT?

Fine by me. :+1:

server/fast_forward.go
```diff
@@ -0,0 +1,223 @@
...
+	mlog.Info("Args", mlog.String("Args", comment))
+	var force bool
+	if len(args) >= 2 {
+		force = args[1] == "--force"
+	}
+
+	backupBranchName := cloudBranchName + "-" + time.Now().Format("2006-01-02") + "-backup"
+
+	for _, repo := range s.Config.CloudRepositories {
+		refs, _, err := s.GithubClient.Git.ListMatchingRefs(ctx, s.Config.Org, repo, &github.ReferenceListOptions{
```
agnivade, line 78: Out of scope for this PR, but do we need to worry about increasing amount of refs as time passes? Wondering if we should delete the older branches to keep the number of refs constant?

agnivade (approved):
Nothing blocking. Nice work on this!

store/sql_lock_store.go
```diff
@@ -0,0 +1,302 @@
...
+	"context"
+	"database/sql"
+	"errors"
+	"fmt"
+	"math/rand"
+	"sync"
+	"time"
+
+	ms "github.com/go-sql-driver/mysql"
+	"github.com/mattermost/mattermost-server/v5/mlog"
```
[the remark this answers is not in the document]
isacikgoz: Sure, created a ticket: https://mattermost.atlassian.net/browse/MM-43287

PR #308, opened 2022-05-05
E2E: Fix wait for loop by removing it, because we use combined status check.

metanerd:
#### Summary
There seem to have been some problems with the revert last time. This should fix it :crossed_fingers:

### DIFF
--- a/server/e2e_test_command.go
+++ b/server/e2e_test_command.go
@@ -4,10 +4,11 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"github.com/mattermost/mattermost-mattermod/model"
-	"github.com/mattermost/mattermost-server/v5/mlog"
 	"net/http"
 	"strings"
+
+	"github.com/mattermost/mattermost-mattermod/model"
+	"github.com/mattermost/mattermost-server/v5/mlog"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


server/e2e_test_command.go
```diff
@@ -93,21 +91,6 @@ func (s *Server) handleE2ETest(ctx context.Context, commenter string, pr *model.
...
-		ghStatusContext = s.Config.E2EServerStatusContext
-	}
-
-	ctx2, cancel := context.WithTimeout(context.Background(), defaultE2ETestTimeout*time.Second)
-	defer cancel()
-	err = s.waitForStatus(ctx2, pr, ghStatusContext, stateSuccess, 30*time.Second) // 30 seconds to consider GitHub's secondary rate limit
-	if err != nil {
-		return err
-	}
-
```
saturninoabril, line 110: Okay, I see, since success checks were already done in the [above LoC](https://github.com/mattermost/mattermost-mattermod/pull/308/files#diff-7047d41fe350835623854ec21ede87cc3c32f0f8842eba189be55bfb1d459bc6L86):
```go
isCIPassing, err := s.areChecksSuccessfulForPR(ctx, pr)
```

saturninoabril (approved):
Looks good to me, thanks @metanerd!

server/e2e_test_command.go
metanerd: I am still not sure, why I cannot query for the single docker build status. But atm figuring this out is not as pressing.

PR #309, opened 2022-05-09
Add Github Organisation Member Refresh for new Members

pfltdv:
#### Summary
Mattermod only fetchs GitHub Organisational Members at startup and there is no refresh mechanism inside. However Mattermod-jobserver fetches members everyday. This causes redeployment of mattermod for new Github members.  

This PR contains cron job to refresh members everyday for mattermod.

#### Ticket Link
Jira: https://mattermost.atlassian.net/browse/DOPS-971

### DIFF
--- a/cmd/mattermost-mattermod/main.go
+++ b/cmd/mattermost-mattermod/main.go
@@ -74,11 +74,7 @@ func main() {
 
 	c := cron.New()
 
-	_, err = c.AddFunc("0 2 * * *", func() {
-		time.Sleep(time.Duration(5) * time.Second)
-		s.RefreshMembers()
-	})
-
+	_, err = c.AddFunc("10 2 * * *", s.RefreshMembers)
 	if err != nil {
 		mlog.Error("failed adding RefreshMembers cron", mlog.Err(err))
 	}
--- a/cmd/mattermost-mattermod/main.go
+++ b/cmd/mattermost-mattermod/main.go
@@ -13,6 +13,7 @@ import (
 	"github.com/mattermost/mattermost-mattermod/metrics"
 	"github.com/mattermost/mattermost-mattermod/server"
 	"github.com/mattermost/mattermost-server/v5/mlog"
+	"github.com/robfig/cron/v3"
 	"golang.org/x/net/context"
 )
 
@@ -70,6 +71,15 @@ func main() {
 			return
 		}
 	}()
+
+	c := cron.New()
+
+	_, err = c.AddFunc("0 2 * * *", s.RefreshMembers)
+	if err != nil {
+		mlog.Error("failed adding RefreshMembers cron", mlog.Err(err))
+	}
+
+	c.Start()
 
 	sig := make(chan os.Signal, 1)
 	signal.Notify(sig, os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
--- a/cmd/mattermost-mattermod/main.go
+++ b/cmd/mattermost-mattermod/main.go
@@ -74,7 +74,11 @@ func main() {
 
 	c := cron.New()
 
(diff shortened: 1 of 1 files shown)


cmd/mattermost-mattermod/main.go
```diff
@@ -71,6 +72,15 @@ func main() {
 		}
 	}()
 
+	c := cron.New()
+
+	_, err = c.AddFunc("0 2 * * *", s.RefreshMembers)
```
phoinixgrr, line 77: Should we run this at a different time (before or after) jobserver cron

https://github.com/mattermost/mattermost-mattermod/blob/54f85a29c3cc53a961862ff31f35427c5276e98a/cmd/jobserver/main.go#L69

pfltdv: I picked same hours to have a same member list in `mattermod-jobserver` and `mattermod` almost at the same time. I do not know how will the applications behave if they have different set of users.

agnivade (approved):

phoinixgrr (approved):

spirosoik:
lgtm but overall it will be a good to have someone else to cross check the code. @agnivade @lieut-data any feedback here

agnivade:
Yep, this is correct. Maybe good to add a slight sleep jitter within 5 seconds to prevent hogging the rate limit quota, but otherwise good to go.

pfltdv:
@spirosoik @agnivade Thanks alot i will add sleep jitter for 5 seconds.

cmd/mattermost-mattermod/main.go
```diff
@@ -74,7 +74,11 @@ func main() {
 
 	c := cron.New()
 
-	_, err = c.AddFunc("0 2 * * *", s.RefreshMembers)
+	_, err = c.AddFunc("0 2 * * *", func() {
+		time.Sleep(time.Duration(5) * time.Second)
```
agnivade, line 78: We need to add the jitter part, which means `rand.Intn(6) * time.Second`. Always sleeping for 5 second means the same thing. We want to spread the load between the two services so that rate limits are not eaten up at the same time.

phoinixgrr: Q: Why don't just run at different timstamps, using different crons for the two services? 10 minutes or 1 hour apart for example. Wouldn't that make it more safe for API rate limiting?

pfltdv: I've tested code little bit more, when `jobserver` and `mattermod` has different set of users, that will cause no issue. Jobserver uses Org Members data only mark contributor PR's with stale label when conditions are met. Since there will be no issue, i am also in favor of starting mattermod cron job later ie. `10 2 * * *`.

pfltdv: Configured mattermod cron job 10 minute later than jobserver.

agnivade: Yep, agree with Akis.

agnivade (approved):
Thanks!

cmd/mattermost-mattermod/main.go
```diff
@@ -71,6 +72,15 @@ func main() {
...
+
+	_, err = c.AddFunc("10 2 * * *", s.RefreshMembers)
+	if err != nil {
+		mlog.Error("failed adding RefreshMembers cron", mlog.Err(err))
+	}
+
+	c.Start()
+
 	sig := make(chan os.Signal, 1)
 	signal.Notify(sig, os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
```
lieut-data, line 86: For completeness, should we `c.Stop()` after shutting down?

pfltdv: c.Stop() is not called also for jobserver. I will create a followup ticket to call Stop after shutdown.

PR #316, opened 2022-06-29
Fix master breakage

agnivade:
We need to set the value of EnableColor field.
Otherwise it leads to a nil pointer dereference
when constructing the logger.

### DIFF
--- a/config/config-mattermod.default.json
+++ b/config/config-mattermod.default.json
@@ -112,6 +112,7 @@
     "LogSettings": {
         "EnableConsole": true,
         "ConsoleLevel": "DEBUG",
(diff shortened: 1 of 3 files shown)


isacikgoz (approved):
🙈

isacikgoz:
Did mattermod deployed after the dependency bump?

agnivade:
> Did mattermod deployed after the dependency bump?

Nope

agnivade:
Oh man GitHub has a new bug. Ctrl-Enter closes the issue rather than just commenting. And it's in my muscle memory so I am repeatedly closing the PR.

PR #317, opened 2022-06-30
MM-43288: Cleanup old cloud branches

agnivade:
We create a cron job that will once a month
and clean up old branches.

Also fix an unrelated bug with resp.NextPage.
We need to check for NextPage == 0, _after_
processing the current batch, not before. This
skips the last batch of items.

https://mattermost.atlassian.net/browse/MM-43288


isacikgoz (approved):
LGTM 👍

PR #318, opened 2022-07-29
E2E: Fix checking for same environment variables.

metanerd:
#### Summary
 Fixes https://github.com/mattermost/mattermost-webapp/pull/10857/#issuecomment-1199233250 edge case. 
I noticed that build tag was wrong, but it also seems, we set it when preparing the test cycle. Do we even need it? https://git.internal.mattermost.com/qa/cypress-ui-automation/-/jobs/984542#L2097 e.g.

### DIFF
--- a/server/gitlab.go
+++ b/server/gitlab.go
@@ -191,17 +191,17 @@ func hasSameEnvs(info *E2ETestTriggerInfo, glVars []*gitlab.PipelineVariable) (b
 	}
 
 	// Check: if the options are not exactly equal, it's not the same pipeline
-	for requiredVar, _ := range glRequiredEnvVars {
+	for requiredVar := range glRequiredEnvVars {
 		delete(glEnvVars, requiredVar) // It's fine even if keys that are not there
 	}
 	if len(glEnvVars) != len(info.EnvVars) {
 		return false, nil
-	} else {
-		for envVar, envVarValue := range info.EnvVars {
-			glEnvVarValue, glEnvVarExists := glEnvVars[envVar]
-			if !glEnvVarExists || glEnvVarValue != envVarValue {
-				return false, nil
-			}
+	}
+	// There's an equal number of options, check if there's any difference
+	for envVar, envVarValue := range info.EnvVars {
+		glEnvVarValue, glEnvVarExists := glEnvVars[envVar]
+		if !glEnvVarExists || glEnvVarValue != envVarValue {
+			return false, nil
 		}
 	}
 


server/gitlab.go
```diff
@@ -161,6 +165,11 @@ func hasSameEnvs(info *E2ETestTriggerInfo, glVars []*gitlab.PipelineVariable) (b
 				}
 				i++
 			}
+			// Triggering PR defaults to 6 created default pipeline environment variables.
+			// See triggerE2EGitLabPipeline in gitlab.go . Adding a custom env vars has 7.
+			if len(glVars) > 6 {
+				return false, nil
+			}
```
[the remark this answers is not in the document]
saturninoabril: Btw, I feel like the comparison inside `if info.EnvVars == nil {...}` could be omitted since it's likely caught by the next for-range loop. Could that be replaced by the following?
```go
if len(info.EnvVars) != len(glVars) {
    return false, nil
}
```

saturninoabril (commented):
Thanks @metanerd for fixing? See comments.

> I noticed that build tag was wrong, but it also seems, we set it when preparing the test cycle. Do we even need it?

I'm inclined to handle it in the pipeline directly (cypress-ui-automation) since it's where we are checking out the actual PR branch. Therefore, maybe [this lines](https://github.com/mattermost/mattermost-mattermod/blob/cd2e3126067ad6e34db0c0b34392583530614b42/server/gitlab.go#L71-L73) could be removed?

server/gitlab.go
```diff
@@ -141,37 +139,85 @@ func (s *Server) checkPipelinesForSameEnvs(ctx context.Context, info *E2ETestTri
...
+
+	isSame, err = isPipelineForSamePRWithSameOptions(info, glVars)
+	if err != nil {
+		return false, err
+	}
+	if isSame {
+		return true, nil
+	}
+
+	return false, nil
```
saturninoabril, line 158: Maybe this could be
```suggestion
	return isSame, nil
```

saturninoabril (approved):
Thanks @metanerd! Left last comment but all good to me.

server/gitlab.go
```diff
@@ -141,37 +139,81 @@ func (s *Server) checkPipelinesForSameEnvs(ctx context.Context, info *E2ETestTri
...
+// "value": "c4a1b4cc4a7fdce4e007e95c840e59e0d976e3f3"
+// }
+// ]
+func isPipelineForSamePRWithoutOptions(info *E2ETestTriggerInfo, glVars []*gitlab.PipelineVariable) (bool, error) {
+	defaultTriggerEnvVars := [5]string{
+		"PR_NUMBER",
+		"REF_MATTERMOST_SERVER",
+		"SHA_MATTERMOST_SERVER",
+		"REF_MATTERMOST_WEBAPP",
+		"SHA_MATTERMOST_WEBAPP",
```
mvitale1989, line 191: These are also defined as constants in the same file, is there any reason not to reuse those instead?
Same for in `defaultEnvs` definition

server/gitlab.go
```diff
@@ -180,18 +222,16 @@ func hasSameEnvs(info *E2ETestTriggerInfo, glVars []*gitlab.PipelineVariable) (b
...
 			}
-			if glVar.Key == envKeyBuildTag && glVar.Value != info.BuildTag {
-				return false, nil
-			}
-
 			if k == glVar.Key && v == glVar.Value {
 				i++
+				optsCount := len(info.EnvVars)
+				if optsCount == i {
+					return true, nil
```
mvitale1989, line 229: If the `glVar` array contains all the options before the `PR_NUMBER`, then this line would run, even though the code has never checked if the PR_NUMBER is the same.
Not sure if the code could allow such a case right now, but it's very easy to look over

mvitale1989: Ah I now notice this only happens inf `len(info.EnvVars)` is 1, in any other case the PR_NUMBER check is done at least once in that first loop

mvitale1989: Added the failing test case, and updated the logic to fix that

server/gitlab.go
mvitale1989: noice, thanks!

mvitale1989 (approved):
Updated the logic to solve the edge case with a single option passed (used to bypass PR number checking), LGTM now

saturninoabril (approved):
Thanks both, LGTM!

PR #333, opened 2022-12-12
cloud fast forward: check branch names instead of commit dates

isacikgoz:
#### Summary
Copied over ticket, It seems like there is a weird behavior that causes taking backup if there is no need to. Our decision-making algorithm works like this:

- Get the most recent backup branch
- Check if the  head commit is older than 5 days, If so take a backup
- Delete the cloud branch
- Create the backup
- Create a new cloud branch

In this case, if you run a `/cloud-ff` command, in the first run everything looks fine. But in the second run, it should be idempotent. But if the `master` branch has last commit more than 5 days ago, it's thinking that we haven't backed up that branch yet. Hence, creating new backups. We should look at the branch names here.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-48376

### DIFF
--- a/server/fast_forward.go
+++ b/server/fast_forward.go
@@ -26,7 +26,7 @@ const (
 )
 
 var (
-	backupRegex = regexp.MustCompile(`cloud-20(\d{2})-(\d{2})-(\d{2})-backup`)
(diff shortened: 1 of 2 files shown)


agnivade (approved):
Not sure how I missed this in my earlier review. I always thought this is how we did it.

Don't forget to ask release team for a re-deploy.

PR #334, opened 2022-12-26
[MM-48414] server/cloud-tag: add cloud tag slash command

isacikgoz:
#### Summary
Adds new slash command as `/cloud-tag` to create a tag based on the head commit of the `cloud` branch.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-48414

### DIFF
--- a/server/cloud_tag.go
+++ b/server/cloud_tag.go
@@ -1,0 +1,146 @@
+// Copyright (c) 2017-present Mattermost, Inc. All Rights Reserved.
+// See License.txt for license information.
+
+package server
+
+import (
+	"bytes"
+	"context"
+	"net/http"
+	"strings"
+	"text/template"
+	"time"
+
+	"github.com/google/go-github/v39/github"
+	"github.com/mattermost/mattermost-mattermod/model"
+	"github.com/mattermost/mattermost-server/v6/shared/mlog"
+)
+
+const (
+	cloudTagName = "tags/cloud"
+	dryRunFlag   = "--dry-run"
+	forceFlag    = "--force"
+)
+
+type cloudTagResult struct {
+	Tagged  map[string]string
+	Skipped []string
+	DryRun  bool
+}
+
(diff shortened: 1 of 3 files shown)


server/cloud_tag.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	// Check if the args are correct
+	command := getCloudTagCommand(comment)
+	args := strings.Split(command, " ")
+	mlog.Info("Args", mlog.String("Args", comment))
+	var force, dryRun bool
+	if len(args) >= 2 {
+		force = args[1] == forceFlag
+		dryRun = args[1] == dryRunFlag
```
agnivade, line 68: Let's use the same constants in fast_forward.go as well

server/cloud_tag.go
```diff
@@ -0,0 +1,146 @@
...
+		result.DryRun = dryRun
+	}
+
+	tagName := cloudTagName + "-" + time.Now().Format("2006-01-02") + "-1"
```
agnivade, line 72: Curious, what is the logic behind the "-1". @phoinixgrr - do you know?

agnivade (commented):
A minor comments.

We should also let Amy know that first the cloud branch needs to be created, and then the tag.

server/cloud_tag.go
phoinixgrr: 🤷

agnivade: Is it okay if we remove it?

phoinixgrr: I really have no idea. 
If I had I would had provided some context in the previous question

agnivade: @spirosoik - is it okay if we remove the "-1" at the end of the tag?

isacikgoz: ping @spirosoik

isacikgoz:
Closing as we are already in the process of deprecating mattermod.

PR #339, opened 2023-04-04
Do not modify the user gitconfig

mvitale1989:
#### Summary

Recently, mattermod failed to perform the cherry-pick operation, which seemed to be related to an inconsistency of expectations around repository cloning and user gitconfig management (tldr: mattermod modified the gitconfig at repo clone time; but repositories persist in an EBS volume, while `.gitconfig` gets lost on a Pod restart, leading to a situation where the repo is cloned but the gitconfig is not as expected). Please check [this discussion](https://community.mattermost.com/private-core/pl/xju696o8rbrwbjirng38yf6eue) for details.

This PR changes the way mattermod behaves in regard to the user gitconfig: instead of modifying it at repository clone time, it will instead assume that it's been already configured so that `git` commands will work as expected.

### DIFF
--- a/config/config-mattermod.default.json
+++ b/config/config-mattermod.default.json
@@ -5,7 +5,6 @@
     "GitHubTokenReserve": 10,
     "GithubUsername": "",
     "GithubEmail": "",
-    "GitConfigMergeRenameLimit": "15000",
     "GithubAccessTokenCherryPick": "",
     "GithubWebhookSecret": "",
     "Org": "",
--- a/server/cherry_pick.go
+++ b/server/cherry_pick.go
@@ -367,13 +367,3 @@ func runCommand(cmd *exec.Cmd, dir string) error {
 	)
 	return cmd.Run()
 }
-
-func runCommandWithOutput(cmd *exec.Cmd, dir string) (string, error) {
-	cmd.Dir = dir
-	cmd.Env = append(
-		os.Environ(),
-		os.Getenv("PATH"),
-	)
-	out, err := cmd.CombinedOutput()
-	return string(out), err
-}
--- a/server/cherry_pick.go
+++ b/server/cherry_pick.go
@@ -351,38 +351,6 @@ func cloneRepo(ctx context.Context, cfg *Config, repoName string) error {
 		return err
 	}
 
-	// Set username and email.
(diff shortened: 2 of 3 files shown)


agnivade:
Do we know why is it failing? If there is no .gitconfig, then the command should just set it. Removing this puts the responsibility on the operator of the tool to remember to set .gitconfig.

mvitale1989:
Yes, as explained [here](https://community.mattermost.com/private-core/pl/t4atuxjixtn95ge7anfm8sygoc) the git configuration commands are run in the `cloneRepo` function, but since the repo dirs are persistent while `.gitconfig` is not, we may get into a case where the repo dir exists (implies: `cloneRepo` won't be run on the next cherry-pick) but the .gitconfig file doesn't exist at all. So you'd get an error message such as the one [Doug reported here](https://community.mattermost.com/private-core/pl/xju696o8rbrwbjirng38yf6eue).

An example of how this could happen is a simple restart of the mattermod Pod, which wipes .gitconfig but keeps the repository directories.

agnivade:
> This would be accompanied by an MR to the mattermod deployment repo, in which we inject the .gitconfig file through a ConfigMap

Ah yes, this sounds better.

agnivade (approved):
Looks good, but the configMap deployment needs to happen first.

mvitale1989:
On a second look, my understanding is that the mattermod config in the JSON is used in more places than just the `git` CLI command configuration file, unfortunately (e.g. [here](https://github.com/mattermost/mattermost-mattermod/blob/master/server/cherry_pick.go#L220) and [here](https://github.com/mattermost/mattermost-mattermod/blob/master/server/cherry_pick.go#L387)).

This means that while the global git config ([which has been merged](https://git.internal.mattermost.com/cloud-sre/kubernetes-workloads/gitops-release-devops/-/blob/main/apps/release-prod/helm-values/release/mattermod-values.yaml#L94-99) last week) indeed helps in terms of stability, we still cannot get rid of the git config entries in the mattermod configuration file. So we now have two places to set the gitconfig (global `.gitconfig`, and `mattermod_config.json`).

As this duplication is already source of confusion (and this PR is anyway a no-op if a global git config is given), i'd now be tending towards just adding the global gitconfig without removing these lines, to document the expectation that the global gitconfig matches the mattermod config. Closing this PR

PR #340, opened 2023-04-06
Fix inverted upstream and origin remotes

agnivade:
And also bump the timeout now that
the main repo has increased in size.

### DIFF
--- a/server/cherry_pick.go
+++ b/server/cherry_pick.go
@@ -42,7 +42,7 @@ func (s *Server) listenCherryPickRequests() {
 
 	for job := range s.cherryPickRequests {
 		func() {
-			ctx, cancel := context.WithTimeout(context.Background(), defaultRequestTimeout*2*time.Second)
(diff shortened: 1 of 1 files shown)


toninis (commented):
Are we sure that this will solve the issue and not break functionality ?

agnivade:
Can you elaborate on your concerns? We did this change to fix the issue, so yes I think we should fix the code as well.

toninis:
> Can you elaborate on your concerns? We did this change to fix the issue, so yes I think we should fix the code as well.

My concern is how did it work the first place . Or did it?

toninis (approved):
It's a small change. If there is an issue we revert

isacikgoz (approved):

agnivade:
@toninis - Any thoughts on the above?

toninis:
> @toninis - Any thoughts on the above?

We can try and do this yes . I agree with you . We will take this from here . I created a [task](https://mattermost.atlassian.net/browse/CLD-5529)

PR #347, opened 2023-09-29
CLD-6234 Cherrypick directly to upstream repo

mvitale1989:
#### Summary

This PR changes the cherry-picking script to push the automated cherry pick branches directly to the `upstream` repo, rather than its `origin` (which is assumed to be a fork; in fact this script originate from the [kubernetes community repo](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-release/cherry-picks.md), where contributions are obviously made from forks).

Example usage: [this PR](https://github.com/mattermost/test-private-project/pull/3) has been created by the following script.
```
# NB: for this to work on your machine, ensure that your ssh key is SSO-authorized against the mattermost org
export ORIGINAL_AUTHOR=mvitale1989
export GITHUB_USER=mvitale1989
export GITHUB_TOKEN=xxxxx
#export DRY_RUN=yes
git clone git@github.com:mattermost/test-private-project.git
cd test-private-project

# Add an `upstream` remote
git remote add upstream $(git remote get-url origin)
git fetch upstream
git checkout upstream/master

### Create dummy changes
BRANCHNAME="dummychange-$(date +%s)"
git checkout -b $BRANCHNAME
echo "dummycontent" > dummyfile
git add dummyfile
git commit -m "Add dummyfile"
git push --set-upstream origin $BRANCHNAME

## Create cherrypick PR of the HEAD commit against the master branch
## Make sure that you checkout this PR
../mattermost-mattermod/hack/cherry-pick.sh upstream/master dummyPR $(git rev-parse HEAD)
```

I added a few more infos in [this comment in the ticket](https://mattermost.atlassian.net/browse/CLD-6234?focusedCommentId=156622), since they include a few sensitive links.

#### Ticket Link

https://mattermost.atlassian.net/browse/CLD-6234

### DIFF
--- a/hack/cherry-pick.sh
+++ b/hack/cherry-pick.sh
@@ -57,7 +57,7 @@ function make-a-pr() {
   local rel
   rel="$(basename "${BRANCH}")"
   echo
-  echo "+++ Creating a pull request on GitHub at ${GITHUB_USER}:${NEWBRANCH}"
+  echo "+++ Creating a pull request on GitHub at ${MAIN_REPO_ORG}:${NEWBRANCH}"
 
   # This looks like an unnecessary use of a tmpfile, but it avoids
   # https://github.com/github/hub/issues/976 Otherwise stdin is stolen
@@ -77,7 +77,7 @@ NONE
 
 EOF
 
-hub pull-request -F "${prtext}" -h "${GITHUB_USER}:${NEWBRANCH}" -b "${MAIN_REPO_ORG}:${rel}"
+hub pull-request -F "${prtext}" -h "${MAIN_REPO_ORG}:${NEWBRANCH}" -b "${MAIN_REPO_ORG}:${rel}"
 }
 
 REPO_ROOT="$(git rev-parse --show-toplevel)"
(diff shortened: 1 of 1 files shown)


phoinixgrr (approved):
Awesome! 
First class PR description 🎖️

toninis (commented):
🚀 I approve this change . 
We might have to remove the repositories from the pods but we will see 😄

toninis (approved):

isacikgoz (commented):
I don't have the historical data on why cherry-pick branches were not wanted in upstream (mattermost/mattermost), maybe it was a security concern?

mvitale1989:
Thank you for the reviews!

@isacikgoz I also don't have the insight in the original decision, but the architecture was also slightly different back then (we used to trigger a CircleCI pipeline, instead of a Github Action workflow). We recently discussed with security in [this thread](https://community.mattermost.com/private-core/pl/5rfu5g4sn7bwurd71tuuwic5po) and the preferable solution is to allow mattermod to push to the repos, rather than allowing worklflows to run in forks.

@toninis good point :smile:  although afaik no new cloning will be required, the `origin` and `upstream` remote remain exactly the same. The only thing that changes it that now we use `upstream` remote for both pulling and pushing

isacikgoz (approved):
Thanks!

PR #352, opened 2023-12-15
feat: Remove excess code from mattermod

toninis:
#### Summary

- Complete deprecation of CircleCI integrations
- Complete deprecation of all cloud related interactions since the Synchronised Monthly Releases are in place
- Complete removal of interaction with the `Build Apps for PR` labels  for mobile and desktop repo

#### Related PRs: 
- https://github.com/mattermost/desktop/pull/2930
- https://github.com/mattermost/mattermost-mobile/pull/7718


phoinixgrr:
Is there a reason for specifically identifying and isolating the code associated with each feature we carve-off from Mattermod? 

This means eliminating the code, testing, perform a build, publishing the build, and redeploying. Moreover emoving code may potentially mess with other functionality that have not yet been transitioned to Argo-Events.

Given the intention to phase out Mattermod, in favor of Github-Actions and the argo-events platform, there's no necessity to uphold its maintenance.

toninis:
I think it is important to do this. 
We need to clean up and keep only the components we are using to choose a better strategy for the future . 
This is still far from phasing out and we need to maintain it . 
For me at least it's an interaction with the code to understand where we use it and why in order to provide the best solution possible for migration .

phoinixgrr:
> I think it is important to do this. We need to clean up and keep only the components we are using to choose a better strategy for the future . This is still far from phasing out and we need to maintain it . For me at least it's an interaction with the code to understand where we use it and why in order to provide the best solution possible for migration .

I would propose doing something like: 

1. Start by performing an analysis of the services provided by Mattermod. 
2. Document these services 
3. Keep track of the ongoing migration progress for each service and consistently update the documentation.
4. Repeat 3. until all services are migrated. 
5. Proceed with decommissioning Mattermod once all services have been successfully migrated.

But not strongly opinionated on this. 2/5.  
Makes sense, to do it also the hard way :)

lieut-data (approved):
I'm in favour of removing code that's been replaced: it's very helpful when I'm grepping across repositories to not think I've found the source of something only to realize it's been deprecated for another thing.

devinbinnie (approved):

isacikgoz (approved):

mvitale1989 (approved):

